PR #228, opened 2023-04-24
Revert "[Ush-191] Add Sentry.io"

miraxes:
Reverts ushahidi/platform-client-mzima#227


tuxpiper:
Why? :( @miraxes @Ifycode

PR #244, opened 2023-05-01
docker: add support for debuggable builds and multisite configuration

tuxpiper:
Related to USH-279


tuxpiper:
@miraxes did I capture it right that the `backend_url` setting would be supposed to look like this in a multi-site config scenario?

```
"backend_url": {
        "domain": "ushahidi.io",
        "api_domain": ""   # optional, defaults to api.{domain}
},
```

When trying this locally with the following setup:
- File `.env.docker` with contents: `MULTISITE_DOMAIN=ushahidi.io`
- Added line in /etc/hosts that assigns 127.0.0.1 to tuxpiper.ushahidi.io
- Run `make start`

Trying to access http://localhost and looking at the Network console hoping to see a request going to http://api.tuxpiper.ushahidi.io , I get this in the error console:

```
TypeError: backendUrl.indexOf is not a function
    at Module.checkBackendURL (environment.ts:2:18)
    at env.service.ts:18:24
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (asyncToGenerator.js:3:1)
    at _next (asyncToGenerator.js:25:1)
    at _ZoneDelegate.invoke (zone.js:372:26)
    at Object.onInvoke (core.mjs:26274:33)
    at _ZoneDelegate.invoke (zone.js:371:52)
    at Zone.run (zone.js:134:43)
    at zone.js:1275:36
```

It looks like there's a check in apps/web-mzima-client/src/app/core/services/env.service.ts that assumes that `backend_url` will be a string.

miraxes:
@tuxpiper  I'm sorry, totally forgot about multi site support for loading configs (which remain in main web not SDK) 
Should be fixed now, can you double check it, please.

PR #259, opened 2023-05-07
chore: add sentry environment configuration

tuxpiper:
This is a useful configuration bit to separate different environments in Sentry.

Did I add it properly?


Ifycode:
> Seems good!

I think so too

tuxpiper:
Thank you folks 🙏  Merged.

PR #291, opened 2023-05-25
[USH-340] View Responses (Map)

DmitryFast:


libs/sdk/src/lib/services/posts.service.ts
```diff
@@ -17,14 +17,14 @@ import { ResourceService } from './resource.service';
 })
 export class PostsService extends ResourceService<any> {
   private defaultPostsFilters: GeoJsonFilter = {
-    order: 'desc',
-    orderby: 'created',
-    set: '',
-    order_unlocked_on_top: true,
-    reactToFilters: true,
-    'source[]': [],
+    // order: 'desc',
```
miraxes, line 20: Is this testing values?

DmitryFast: Yes. Filters is not ready yet, so in order to display posts, I need to set them statically.

kvengerov (approved):

PR #316, opened 2023-06-09
Removed Unused Variables

MugoBrian:
Remove unused variables and imports and tests

### DIFF
--- a/apps/web-mzima-client/src/app/app.component.spec.ts
+++ b/apps/web-mzima-client/src/app/app.component.spec.ts
@@ -17,13 +17,6 @@ describe('AppComponent', () => {
     expect(app).toBeTruthy();
   });
 
-  it(`should have as title 'platform-client'`, () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
-
-    expect(app.title).toEqual('platform-client');
-  });
-
   it('should render title', () => {
     const fixture = TestBed.createComponent(AppComponent);
     fixture.detectChanges();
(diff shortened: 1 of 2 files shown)


miraxes:
@MugoBrian Is there any reasons to use forked repo?
I suggest you to use main repo and create branches on it. So all flows will work correctly.

MugoBrian:
I don't have such access permissions to the repository.

Ifycode:
> @MugoBrian Is there any reasons to use forked repo? I suggest you to use main repo and create branches on it. So all flows will work correctly.

@tuxpiper

Ifycode:
@miraxes it's ok for @MugoBrian to remove the title since it is unused right? Or will it be used for anything later on? Because I see some other thing in the code controls/changes the page Tab title.

miraxes:
Its ok to remove it, we can easily add it later in case we need.

Ifycode:
Ok. Thank you.

apps/web-mzima-client/src/app/app.component.spec.ts
```diff
@@ -17,13 +17,6 @@ describe('AppComponent', () => {
     expect(app).toBeTruthy();
   });
 
-  it(`should have as title 'platform-client'`, () => {
-    const fixture = TestBed.createComponent(AppComponent);
-    const app = fixture.componentInstance;
-
-    expect(app.title).toEqual('platform-client');
-  });
-
```
Ifycode, line 26: @MugoBrian Please leave the tests for now since that is a different ball game. We will address this, but not now.
So just add the test code back for now.

Ifycode (changes requested):

apps/web-mzima-client/src/app/app.component.spec.ts
MugoBrian: Ok I will. Actually @kvengerov instructed me to do so.

miraxes: yeah, @Ifycode  tests will fail if you make those changes only in app component, so probably we should adjust test files too.

PR #353, opened 2023-06-30
Changed password length for user registration throw settings

kvengerov:


apps/web-mzima-client/src/app/settings/users/user-item/user-item.component.ts
```diff
@@ -25,6 +25,7 @@ export class UserItemComponent implements OnInit {
   public isMyProfile = false;
   public isDesktop = false;
   public createUserErrors: any[] = [];
+  private passwordLength = 8;
```
miraxes, line 28: lets use CONST for that @kvengerov

miraxes (approved):

PR #372, opened 2023-07-10
[USH-539] Feed view correct refresh posts

miraxes:


kvengerov (approved):

miraxes:
This is related to USH-439, wrong number :(

PR #374, opened 2023-07-10
chore: add ephemeral API server to E2E tests

tuxpiper:
Adds ephemeral API server to end to end tests run in github actions

### DIFF
--- a/.github/workflows/test-and-ship.yml
+++ b/.github/workflows/test-and-ship.yml
@@ -70,13 +70,77 @@ jobs:
     timeout-minutes: 15
     continue-on-error: true
 
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - "33061:3306"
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-ship.yml
```diff
@@ -70,13 +70,77 @@ jobs:
     timeout-minutes: 15
     continue-on-error: true
 
+    services:
+      mysql:
+        image: mysql:5.7
+        ports:
+          - "33061:3306"
+        env:
+          MYSQL_ROOT_PASSWORD: root
```
AmTryingMyBest, line 79: @tuxpiper are these okay?

tuxpiper: yes it's ok @AmTryingMyBest  . This is an ephemeral, non-publicly reachable mysql service

PR #382, opened 2023-07-14
[USH-508] Color contrast accessibility issue

MugoBrian:
[accessibility color contrast](https://linear.app/ushahidi/issue/USH-508/accessibility-color-contrast)
@Ifycode

### DIFF
--- a/libs/mzima-ui/src/lib/button/button.component.scss
+++ b/libs/mzima-ui/src/lib/button/button.component.scss
@@ -169,6 +169,19 @@ $laptop: 1439px;
     &:disabled {
       color: var(--disabled-color);
     }
+
+    &.mzima-button--small {
+      text-decoration: none;
+      padding: 6px 12px;
+      border-radius: 4px;
+      color: var(--color-black);
+      background-color: var(--color-neutral-10);
+
+      &:hover {
+        --hover-background-color: none;
+        background-color: var(--color-net);
+      }
+    }
   }
 
   &:not(.mzima-button--clear) {
--- a/apps/web-mzima-client/src/app/post/post-metadata/post-metadata.component.scss
+++ b/apps/web-mzima-client/src/app/post/post-metadata/post-metadata.component.scss
@@ -64,7 +64,7 @@
     margin: 0 -8px -4px;
     align-items: flex-start;
     justify-content: flex-start;
-    color: var(--color-neutral-70);
+    color: var(--color-neutral-80);
   }
 
   &__item {


libs/mzima-ui/src/lib/button/button.component.scss
```diff
@@ -169,6 +169,19 @@ $laptop: 1439px;
     &:disabled {
       color: var(--disabled-color);
     }
+
```
MugoBrian, line 172: OPTION 1: put this &.mzima-button--small within the --clear styling
    Advantages
    I will not have to add a color property to each clear button in the search-form.component.html
    Disvantages
    Each element with the class --small and --clear will be forced to have this styling.
    OPTION 2: put this &.mzima-button--small within the --gray styling
    Advantages
    I don't have to force this styling to eacrh and every element which will have a class of --small and --clear
    This styling will ONLY appear to elements with the class (property) --small and --clear and --gray 
    Disadvantages
    I will have to add a color property to each clear button in the search-form.component.html
    Which option is better in terms of CAM 2023? Or is there any better option?

Ifycode: One of the purpose of the CAM project is to "untangle". So I'd say that all the suggestions feel like an hack @MugoBrian 
It is already confusing that the --clear does more than take care of "clear".

Ifycode: @DmitryFast @miraxes help me understand how this part works. What is the purpose of `&--clear`?
Asking because it looks like it takes care of many other buttons that do not serve the "clear" purpose (to me).

PS: The styles in this area are very confusing.

Ifycode (changes requested):

libs/mzima-ui/src/lib/button/button.component.scss
MugoBrian: The --clear class applies a transparent background with a #ffc334 color basically to the button. The fill property can have solid, outline or clear.

MugoBrian: > One of the purpose of the CAM project is to "untangle". So I'd say that all the suggestions feel like an hack @MugoBrian It is already confusing that the --clear does more than take care of "clear".

Let me see if I can target the specific element needed.

Ifycode: Also @MugoBrian This is not the issue I'm expecting a PR for as sent to your dm.

Ifycode: Ok. Like I said, I understand that part of the fill types and styles as I have responded to Brian earlier - There is even comment in the code that explains that part. 

Probably I did not explain myself well or ask it properly earlier. so I'm going to do a bit of rephrase.

There are 2 perspectives to my question. 
- One which has to do with the clash of the word "clear" between the style name and the purpose of the clear button.
- The second has to do with the screenshot I have below (this however is even general, not just related to the clear button - but I'll use it as an example):
There were 2 classes used on the clear button `--primary` and  `--clear`. Plus the one added by Brian `--small` making 3 now. All 2 or 3 classes change the colour and background. My problem is this: so what's the essence of having all 2 or 3 classes on one element when (majorly) they all change colour and background? My reason for mentioning earlier that adding the extra class feels like an hack, and also for stating that it is confusing earlier.

I could go further, but I will stop here. As this is the part I feel is related to the discussion in this PR.
Based on all I have said, an advice on how to go about this for Brian will be helpful - One that we can also adopt for other places where this problem occurs.

#
Screenshot 2023-07-19 at 7 48 31 AM

DmitryFast: Oh, I understand what you mean. Yes, I agree, if we have a yellow button that turns black when you add fill="clear" to it, that's not right. In this case, you need to add a new color for the button, or use one of the existing ones, for example "medium". I thought that our button remains yellow, it just becomes a darker shade

PR #443, opened 2023-08-02
Restore data QA input directive earlier removed

Ifycode:
PR restores data QA input directive earlier removed


Ifycode:
@miraxes @AmTryingMyBest Added it, please check it out.

miraxes (approved):
Seems good.

PR #564, opened 2023-09-01
[USH-789] Extra margin appearing in the account component for the laptop view.

MugoBrian:
As per our discussion @Ifycode . I have fixed the margin issue for this [ticket](https://linear.app/ushahidi/issue/USH-789/extra-margin-appearing-in-the-account-component-for-the-laptop-view).

Kindly review it and confirm if the fix works and the e2e tests are passing.


Ifycode (commented):
Don't worry, just live this PR open.
I'll check and do the rest.

PR #593, opened 2023-09-29
[USH-543] Survey creation confirm cancel

miraxes:


miraxes:
@Mh-Asmi I'm not sure if you need to put this in release directly, because this is not Urgent

PR #634, opened 2023-10-11
E2e collection test automation

### LINKED ISSUE #597: Collections
- [ ] Login as an admin
- [ ] Navigate to collections 
- [ ] Create a new collection
- [ ] Click Add Collection
  - [ ] Enter name as: "Automated Collection"
  - [ ] Enter Description as: "Automated Description"
  - [ ] Select "Only me" radio button
  - [ ] Click Save
- [ ] Add post to Collection
- [ ] Verify post added to Collection

shakirandagire:
@AmTryingMyBest Please review this PR which resolves [597](https://github.com/ushahidi/platform-client-mzima/issues/597)


e2e-testing/cypress/functions/CollectionFunctions.js
```diff
@@ -0,0 +1,96 @@
+import CollectionLocators from '../locators/CollectionLocators';
+
+class CollectionFunctions {
+
+    launch_collection_modal(collectionURL) {
+        cy.visit(collectionURL);
```
theobadha: @shakirandagire Let's launch the collections modal through the interface, not directly using a URL
So login, and then navigate into the collections by clicking on the Collections button.

shakirandagire: @AmTryingMyBest Thanks for the feedback, let me do that

shakirandagire:
@AmTryingMyBest I have implemented the feedback on the PR. Could please review again

apps/web-mzima-client/src/app/shared/components/collections/collections.component.html
theobadha, line 1: Good effort to try adding the custom attribute selectors. But this didn't seem to quite work. The custom attributes don't appear on the front-end and thus the test doesn't run to completion. 
Two options at this point:
- You explore this further to see if you can get the custom attributes added right (may require some basic angular knowledge) 
- You ignore adding the custom attributes for the collections, and you use selectors that will work. 
If you choose option one I will create a separate task for that so that that effort is acknowledged.
LMK! :)

shakirandagire: Thank you for the feedback.
I choose option one which requires exploring this further because it is a learning opportunity. Please let me know when you create the separate task for this.

theobadha: Actually, please wait. This may be right after all. I'll review this a bit more closely later this afternoon and will get back.

theobadha: @shakirandagire the code works fine. I was pointing into the wrong environment. 
Well done! 👍

shakirandagire: Thank you for the feedback @AmTryingMyBest

PR #638, opened 2023-10-12
[USH-870] the "+" sign not recognized as a valid character in logging in email field

Ifycode:
Fixes [USH-870] the "+" sign not recognized as a valid character in logging in email field


Ifycode:
@Mh-Asmi @webong PR is ready for review

Issue #607, opened 2023-10-04
Users


theobadha:
- [ ] Delete more than one user

WanjaNjunge:
@AmTryingMyBest I'd like to pick this up. Please assign it to me.

PR #643, opened 2023-10-17
Done with adding Delete more than one user test suite

WanjaNjunge:
@AmTryingMyBest can you kindly review this PR for issue Users #607


e2e-testing/cypress/functions/UserFunctions.js
```diff
@@ -34,6 +36,36 @@ class UserFunctions {
...
+    }
+  
+    this.open_user_page_steps();
+    cy.get(UserLocators.bulkActionsBtn).click();
+    userNames.forEach((userName, index) => {
+      // Type the user name in the search input
+      cy.get('.p-inputtext').clear().type(userName);
+
+      // Select the checkbox next to the user
+      cy.get(`:contains("${userName}") .p-checkbox-box`).eq(0).click();
```
theobadha, line 52: For this to work, we'd need users with the names to exist so that on searching, there are results to be found and acted on.
For this test to run correctly, we can do one of two things:
 - create the users within the test, then find and delete them
 - run the test against users that already exist in the environment. 
I'll hold off merging this until we can sort this out.

WanjaNjunge: Okay. Let me look into that. Thanks for the feedback

WanjaNjunge: I have opted to run the tests against users that already exist and committed the changes. @AmTryingMyBest your review will be appreciated. Also noticed that the first test, Adding a user fails. Can I fix it under this PR?
Screenshot from 2023-10-23 13-58-14

PR #680, opened 2023-10-31
chore: add linting cache locally and to actions

tuxpiper:
So much faster with caching. I would have never imagined that linting could be so resource-intensive.


webong (approved):
This brings lots of positive improvement to the CI pipeline

PR #708, opened 2023-11-16
[USH-581] Follow Up - Endpoints to manually update Profile Details on mobile app

MMwandigha:
Updated the frontend code for image upload/delete feature
Missing is the backend implementation


Ifycode:
Reviewing @MMwandigha

Ifycode (changes requested):
Nice start @MMwandigha 
Continue by studying the other profile information that can save, see how they were implemented and possibly use the idea to save to the backend.

MMwandigha:
@Ifycode Noted. Let me get into it

MMwandigha:
@Ifycode You can now fully review this

apps/mobile-mzima-client/src/app/profile/information/components/profile-photo/profile-photo.component.ts
```diff
@@ -1,21 +1,135 @@
...
+  }
+
+  changePhoto(file: File): void {
+    const reader = new FileReader();
+    console.log(reader);
+    console.log(file);
+    reader.onload = () => {
+      this.photo = reader.result as string;
+
+      this.mediaService.uploadFile(file, 'Test caption').subscribe((response: any) => {
```
ushahidlee: Should probably change 'Test Caption'!

ushahidlee (changes requested):
I was able to change my profile picture once and then I couldnt change it after that. 

Also, the behaviour is a bit unclear.: after I click "Set new photo" and choose the new picture, it updates in the UI but the "save changes" button is inactive.

I think the issues here are related to events.

apps/mobile-mzima-client/src/app/profile/information/components/profile-photo/profile-photo.component.ts
MMwandigha: Got it. Should we name it "Profile Picture" instead?

MMwandigha:
@ushahidlee Hmm, that's strange.
Oh yeah, I hadn't taken note of the save button. Let me see what I can come up with.

apps/mobile-mzima-client/src/app/profile/information/components/profile-photo/profile-photo.component.ts
ushahidlee: Maybe the user's display name?

MMwandigha: Yeah, I think this may work

MMwandigha:
So here is my attempt at this ticket trying to achieve the following objectives (as shown by the design)

-Implementing the spinner to display once a profile picture is uploading
-Displaying the delete button only in instances where the user has edited the photo before. The delete button should be disabled as long as the default gravatar is shown
-Showing a toast message upon successful deletion

I pushed this code for extensive review before clean up by @ushahidlee , @Ifycode or @Angamanga . (We can arrange to meet). If you have any suggested changes, I can do them before pushing this code again

ushahidlee (changes requested):
Code looks good and the functionality works, but I do not see the spinner while the picture is uploading. Otherwise its great!

MMwandigha:
Thanks @ushahidlee I see...can you try with an image with a darker colors, perhaps the spinner was blending in with white photos. Also I think I should consider changing contrast based on the image color.

MMwandigha:
This PR shows my progress towards optimizing the profile picture upload ticket. Here's what happened

- The user's custom profile picture will only be displayed after the photo has been fully uploaded to the server
- Improved the spinner's visibility by adding a stroke to the spinner. The border property didn't quite work because the spinner changes to white in dark mode and black in light mode

Pending: The profile picture is taking long to upload (approx 30 sec.) if you choose large images. @Angamanga is there a way to mitigate this?

MMwandigha:
To prevent concurrent uploads, I created an emitter that blocks backward navigation as long as the photo is still uploading and also a toast message to tell the users that the photo is still uploading. @ushahidlee Let me know if it works well.

ushahidlee (approved):
Code looks good and works well!

PR #723, opened 2023-12-11
Fix delay in UI while permissions are loading

Angamanga:
This PR adds a spinner the permissions-list when editing or creating a role. 
How to test:
- Logged in as a user with role-creating permissions
- Go to settings, click on roles
- Click on a role or create a new one
- [ ] Spinner appears while the permissions are fetched from the API

Fix for issue https://linear.app/ushahidi/issue/USH-856/role-fields-appear-empty-for-a-sec-and-then-populated-with-data#comment-6cd75721


Ifycode (commented):
Ok, nice idea that the spinner should only affect some content. The code looks ok. I'm just going to check and interact with the UI in a few minutes @Angamanga

Ifycode:
Ok, so the flicker still happens - i.e. the input fields are empty briefly (and for admin and member roles, the delete button also appears briefly).
This is the part where you were referring to for state management right? @Angamanga

Angamanga:
> Ok, so the flicker still happens - i.e. the input fields are empty briefly (and for admin and member roles, the delete button also appears briefly). This is the part where you were referring to for state management right? @Angamanga

Ahh, yes you are true, that is also affected by the delay :/ Good find! I will update the spinner to cover it all until the data is there 👍

Angamanga:
@Ifycode I have updated the PR so the spinner covers the whole form 👍

Ifycode (approved):
This works to hide the flicker for now. So... 👍🏼

PR #726, opened 2023-12-12
Add Mobile Data Source Label

webong:
- Add mobile as a data source to search helper file

### DIFF
--- a/apps/web-mzima-client/src/app/core/helpers/search-form.ts
+++ b/apps/web-mzima-client/src/app/core/helpers/search-form.ts
@@ -20,6 +20,18 @@ export const statuses = [
 
 export const sources = [
   {
+    name: 'Web',
+    value: 'web',
+    total: 0,
+    checked: true,
+  },
+  {
+    name: 'Mobile',
+    value: 'mobile',
+    total: 0,
+    checked: true,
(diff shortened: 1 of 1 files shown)


webong:
@tuxpiper @Mh-Asmi 

This PR needs to go out for the ability to see Mobile Source Label in the Search Bar

Angamanga (approved):
This works fine! Approved. 

Question though, should there be a mobile-label in the app too? I see there is a similar file in mobile-mzima but no "Mobile"-source added there.

tuxpiper:
> This works fine! Approved.
> 
> Question though, should there be a mobile-label in the app too? I see there is a similar file in mobile-mzima but no "Mobile"-source added there.

I would definitely think so, this bit of functionality should be equivalent both on the web and mobile app! Nice catch @Angamanga  ! 
Could you please create a sub-issue  in https://linear.app/ushahidi/issue/USH-987/add-mobile-app-as-a-data-source-label-at-the-backend  ? Urgent priority .

Angamanga:
Sub-issue added: https://linear.app/ushahidi/issue/USH-1085/add-mobile-app-as-a-data-source-label-in-mobile-client-code

PR #729, opened 2023-12-14
[USH-1029] Web Client: Save button is disabled & can not save form after edit - General settings

Ifycode:
[USH-1029] Web Client: Save button is disabled & can not save form after edit - General settings

### DIFF
--- a/apps/web-mzima-client/src/app/settings/general/general.component.html
+++ b/apps/web-mzima-client/src/app/settings/general/general.component.html
@@ -90,7 +90,10 @@
       </div>
       <div class="checkbox">
         <!-- FIXME: isDisableRegistrationEnabled -->
-        <mat-checkbox [data-qa]="'site-disable-registration'" formControlName="disable_registration">
+        <mat-checkbox
+          [data-qa]="'site-disable-registration'"
+          formControlName="disable_registration"
+        >
           {{ 'settings.site_disable_registration' | translate }}
         </mat-checkbox>
         <p translate="settings.site_disable_registration_desc"></p>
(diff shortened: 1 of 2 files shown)


apps/web-mzima-client/src/app/settings/general/general.component.html
```diff
@@ -90,7 +90,10 @@ <h1 *ngIf="isDesktop$ | async">{{ 'settings.settings_list.general' | translate }
       </div>
       <div class="checkbox">
         <!-- FIXME: isDisableRegistrationEnabled -->
-        <mat-checkbox [data-qa]="'site-disable-registration'" formControlName="disable_registration">
+        <mat-checkbox
+          [data-qa]="'site-disable-registration'"
+          formControlName="disable_registration"
+        >
           {{ 'settings.site_disable_registration' | translate }}
```
Ifycode, line 97: For this unrelated change, it keeps rearranging this line at the point of commit or push. This is the first time I'm experiencing this, and I haven't changed my editor settings - so it's a bit confusing.
It also doesn't allow me to change it back via another commit , it will just write this in the terminal:
````
⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration
  ````
  
  I deleted the branch and forced push a new change that does not rearrange this part. it still ends up changed in the file changes of the PR @Angamanga

Angamanga: No worries @ifycode, it is just a formatting-change so lets just keep it 👍

Angamanga (approved):
This works fine and the settings are possible to save 👍 You could consider doing parseFloat on value directly if you want to make that change, but this works and covers the error we had :)

PR #732, opened 2023-12-19
Fixes issue with deleting a post while looking at it in data view.

ushahidlee:
- Added deleted post state to event bus
- Pick up delete event and republish through event bus.

### DIFF
--- a/apps/web-mzima-client/src/env.json
+++ b/apps/web-mzima-client/src/env.json
@@ -1,6 +1,6 @@
 {
   "production": true,
-  "backend_url": "https://mzima-dev-api.staging.ush.zone/",
+  "backend_url": "http://localhost:8080/",
   "api_v3": "api/v3/",
   "api_v5": "api/v5/",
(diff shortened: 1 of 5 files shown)


apps/web-mzima-client/src/env.json
```diff
@@ -1,6 +1,6 @@
 {
   "production": true,
-  "backend_url": "https://mzima-dev-api.staging.ush.zone/",
+  "backend_url": "http://localhost:8080/",
```
Angamanga, line 3: Is this change intentional?

Angamanga (approved):
Works all fine! Just a question, is the change of backend_url intentional? It does make sense since that is the url for the API locally, right?

apps/web-mzima-client/src/env.json
ushahidlee: Good spot, that was an oversight!

Angamanga: Although I think it does makes sense, that is the address the API get when setting it up locally, isn't it? It makes it easier for new contributors to get started.

PR #733, opened 2023-12-19
Deleting a user reverts the users list to the first page

ushahidlee:
- Added a little calculation to make the first list item respect the current page.


Angamanga (approved):
Works fine 👍

PR #735, opened 2024-01-05
940 login sign up UI

Angamanga:
This PR adjusts the UI for the login-flow, [USH-940](https://linear.app/ushahidi/issue/USH-940/update-loginsignup). Designs are here: https://www.figma.com/file/c0LFvzuib8E3FHmp6Of8MS/Ushahidi?type=design&node-id=10059-94985&mode=design

**What is changed?**
- Login-form
- Sign-up form
- Reset password forms
All are available from main-navigation-bar ->Login/signup.

Note: This is a completely new form-style compared to what we use anywhere else. I put the css in the same file as the rest of the form-styles but added an extra class. Is this the way we want to do it? I reasoned it was better to put it there in case we want to reuse it somewhere else but also to avoid repeating lots of code.

I found 2 bugs that I will add new issues for to not make this PR too big:
1. The cancel-button disappears in mobile mode. This has to do with the modal I think and not the form itself.
2. The modal covers the whole screen in mobile mode. This happens for all modals that is opened from the navbar to the left.


ushahidlee (approved):
All looks good, but you have removed the .form-error class and then apply it to a new mat-error tag. Not sure if thats intentional?

Angamanga:
> All looks good, but you have removed the .form-error class and then apply it to a new mat-error tag. Not sure if thats intentional?

Good find, it is actually not used anymore! Will remove 👍

tuxpiper:
@Angamanga  is it ready to merge?

PR #739, opened 2024-01-08
Check and update UI if a Tweet is unavailable/deleted

Angamanga:
Fixes https://linear.app/ushahidi/issue/USH-938/fix-the-spinner-of-deleted-tweets-fe.
To test: In a deployment with incoming Tweets, make sure that:
- The spinner stops spinning 
-  UI is updated and informs user that the Tweet is deleted.


apps/web-mzima-client/src/app/post/twitter-widget/twitter-widget.component.ts
```diff
@@ -35,17 +37,23 @@ export class TwitterWidgetComponent implements OnInit {
...
+
+      /* The Twitter-widget does not resolve the promise if 
+      the Tweet is deleted, so need to check it with a timeout */
+      setTimeout(() => {
+        this.isTwitterScriptLoading = false;
+        if (!this.tweet) {
+          this.isTwitterFailed = true;
+          this.loadingFailed.emit('Tweet failed to load');
+        }
+      }, 30000);
```
Angamanga, line 56: This felt like an awful long time when checking on my Internet-connection, but better to keep it this way to cover users with slow connections.

Ifycode: 🤔

Ifycode: So I'm probably going to leave this for @ushahidlee or other persons on the the team to take a look at and approve.

Something in me was hoping `twitter-wjs` could have some check for if the a tweet still exists, and that it can be checked before or around when the script/element is appended to the DOM: https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/set-up-twitter-for-websites

I have not noticed that it does that from the docs. I was hoping that could happen and we can get to customise our own content - more like instead of twitter-wjs "Not found" & our "tweet unavailable" image showing up at the same time, we could just replace their Not found with our image. Also because it's annoying to see the "Not found" show up at the same time with the loader/spinner, before "tweet unavailable" takes over.

I also wish `twitter-wjs` provided a loading state status that we can just tap from. I'm not sure if that exists, but I feel this would make things cleaner.

Those are just "things I tried to try out" on my own , not like they are a proposed solution.

If nothing else anyone suggests/tries works out, then I guess the set timeout method will just have to suffice.

Angamanga: @Ifycode I tried that but I get no response or indication from the widget when the tweet is missing, its super-annoying and it feels like I have tried everything to detect it. It seems like it is on the twitter-side (if I haven't missed something but I do have looked for it 😅 ). This solution is not elegant but its really the only thing I could come up with. However, I added a small change to remove the spinner immediately when a tweet in fact is there, so a small improvement 👍

ushahidlee: I tried binding to the rendered event as described in the documentation ((https://developer.twitter.com/en/docs/twitter-for-websites/javascript-api/guides/javascript-api) but the event never fires. From what I can see on the twitter forums, stack overflow, etc these events dont work for anyone else either.

ushahidlee (approved):
The best solution available, no thanks to you Elon!

Ifycode (approved):
@ushahidlee 😂
I agree. Good to go 👍🏼

PR #743, opened 2024-01-18
Add conversation-UI to sms-posts (messages)

Angamanga:
Adding the conversation-window to posts originating from sms.

Designs are here: https://linear.app/ushahidi/issue/USH-1077/design-messages-list-and-interaction-user-interface

Note:
1. Left to do: Pagination is not done entirely like the designs because I had problems with the pagination-window.
2. After discussions with Yurii, I removed the modal for sending a new message and added the text-area to the conversation-window instead
3. To test in the browser, you need to look at a post originating from sms, either structured or unstructured.
4. Left to do: No unit-tests added at this point 😢


ushahidlee (changes requested):
Everything looks good and works fine, except for the pagination. When I get to the last page, I can still select "next page" which takes me to a page with no results. As there are no results, this removes the pagination buttons so I cannot return to the previous.

Angamanga:
@ushahidlee Pagination and message-limit updated 👍

ushahidlee (approved):
Works great!

PR #744, opened 2024-01-19
"Members" should be able to delete their own contact

ushahidlee:
Creates the ability for users to switch contact methods if they try to delete a contact method that they receive notifications on. There is functionality in the designs which is out of scope for the ticket, and I have moved into a new ticket. (https://linear.app/ushahidi/issue/USH-1104/how-we-handle-notifications-should-be-revised-based-on-designs)

Associated Tickets: 
 
https://linear.app/ushahidi/issue/USH-941/update-delete-contacts
https://linear.app/ushahidi/issue/USH-888/members-should-be-able-to-delete-their-own-contact


Angamanga (approved):
Looks good and works fine 🎉

PR #747, opened 2024-01-24
1070 hidden field indicator

Angamanga:
This PR adds a lock-icon to the labels in the survey-forms when a field is private.

Issue: https://linear.app/ushahidi/issue/USH-1070/no-indicator-for-what-fields-in-a-survey-that-are-private


Angamanga:
@ushahidlee Do you have time to review today and we might be able to send it out together with the sms-fix 🙏

ushahidlee (approved):
Clean and effective. 5/5 Would review again

PR #748, opened 2024-01-24
Pulling the deployment name and adding it to the sender-field in the UI

Angamanga:
A small change after QA, adding the Deployment-name as the sender to the UI. Previously, the same number was visible both as sending and receiving number by mistake.


ushahidlee (approved):
:+1:

PR #749, opened 2024-01-24
Ush 1107 - New deployment opening map with a zoom level of 0 which suppresses the map

ushahidlee:
Issue: 

When a new deployment is opened for the first time, the map zoom level is set to 0. This effectively hides the map.

Fix:

I have changed the minimum zoom level from 0 to 1, which should prevent the map from disappearing. I have also modified a line that was forcing a zoom level of 0 when no bound was available, to use either the configured default zoom, or the min zoom (1) if default zoom is unavailable.


Angamanga (approved):
All fine, no fuss! Approved!

PR #750, opened 2024-01-25
Re-enabling Intercom on Android

ushahidlee:
Issue:
To resolve crashes reported in USH-1052, Intercom was disabled in the android build. The crashes were related to non-logged in users having access to intercom, and the proguard rules used during build.

Resolutions:
- Intercom reenabled.
- Proguard rules modified so Minification and Obfuscation are enabled, but Shrinking is disabled.
- Intercom is removed as an option for users that are not logged in.

Related Tickets:
https://linear.app/ushahidi/issue/USH-1052/crashes-with-android-build-22023500
https://linear.app/ushahidi/issue/USH-1075/update-intercom-android-sdk-within-ionic-capacitor-intercom-package


apps/web-mzima-client/src/env.json
```diff
@@ -1,6 +1,6 @@
 {
   "production": true,
-  "backend_url": "http://localhost:8080/",
+  "backend_url": "https://mzima-dev-api.staging.ush.zone/",
```
tuxpiper, line 3: I think I have seen this line changing back and forth in a couple PRs , maybe we should just .gitignore it 🤷

tuxpiper (approved):
Looks good to me, just a small comment here: https://github.com/ushahidi/platform-client-mzima/pull/750/files#r1469688657 to take into account somewhere else

PR #754, opened 2024-01-27
E2e fix posts tests

theobadha:
I made quite a few changes in this PR. 
- removed the data-qa="null" selectors and replaced them with a correct selector. I noticed that data-qa="" would end up with the null value, while ng-reflect-data-qa="" would have the correct value. Idk why this happens, but to clean things up in the future we may want to correct that so that all selectors are the same. 
But for now, this works correctly. :) 
 - i removed the step that interacted with the date field. I saw this didn't add value to the test as it was. To get value out of that check, we need to verify the content of the field.


shakirandagire (approved):
The changes look good to me.
The tests pass as expected and the work around for the data-qa field makes sense.

PR #755, opened 2024-01-27
Checking for manage-post permissions before showing sms-conversation

Angamanga:
This PR hides the conversation-window for users without manage-posts permissions.

In both data and map-view, look at a published post coming from sms. When logged out, the conversation-window should not be visible and the user can look at the post details, both in a modal in the map and in dataview. If logged in as an admin or user with manage-post-permissions, the conversations-window is visible.


tuxpiper:
I am going to merge it in order to release this as a hot fix. @webong or @ushahidilee you still owe @Angamanga a review ;)

Issue #720, opened 2023-12-05
Log in with newly created user


theobadha:
We have a test that created a user.
- [x] Let's log in with this user and verify that they are able to successfully log in after they are created.

PR #758, opened 2024-02-01
E2e test for logging in with newly created user

shakirandagire:
@AmTryingMyBest Could you help me review this PR that resolves issue[720](https://github.com/ushahidi/platform-client-mzima/issues/720).
Please note that it has also resolved the failing test for user-signup that had been failing for a while.


e2e-testing/cypress/functions/SignupFunctions.js
theobadha, line 1: It may not be a very good idea to use `{force: true}` with every cypress command. Let's go through these files and remove them where unnecessary.
I created a new issue for it.

shakirandagire: @AmTryingMyBest Alright let me get the alternative to this because Cypress had given me this as a solution for the parent element having visibility hidden.

theobadha: @shakirandagire we can use it only when it's necessary. Not on each element or cypress command. Delete a few and run the tests, see which ones are not required.

shakirandagire: Alright. Let me do that.

theobadha: After reviewing, I'll merge this as is. 
It seems to pass consistently and accurately.

PR #762, opened 2024-02-09
[USH-1109] Make role name uneditable where attribute is protected

Ifycode:
**Frontend fix (Details):** 
- On Edit/Update: Only the display_name now changes - for any role whether protected or not protected. name property does not change.
- On Add: Both display_name and name properties change.

### DIFF
--- a/apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
+++ b/apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
@@ -54,7 +54,6 @@ export class RoleItemComponent implements OnInit {
       description: [''],
       permissions: [[], [Validators.required]],
       id: [null],
-      name: ['', [Validators.required]],
       protected: [false],
     });
   }
@@ -99,7 +98,6 @@ export class RoleItemComponent implements OnInit {
       id: role.id,
       display_name: role.display_name,
       description: role.description,
-      name: role.name,
       protected: role.protected,
       url: role.url,
       permissions: role.permissions,
@@ -119,7 +117,6 @@ export class RoleItemComponent implements OnInit {
     this.isFormOnSubmit = true;
     const roleBody = {
       id: this.form.value.id,
-      name: this.form.value.name,
       display_name: this.form.value.display_name,
       description: this.form.value.description,
       permissions: this.form.value.permissions,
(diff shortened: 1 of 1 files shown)


apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
```diff
@@ -136,6 +134,7 @@ export class RoleItemComponent implements OnInit {
         },
       });
     } else {
+      delete roleBody.name; // We don't want to change name property on edit/update
```
ushahidlee: According to the discussion yesterday we will be requiring the name field to be intact in the update request and that it be unchanged from the original object.

ushahidlee (changes requested):
See line comments

apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
Ifycode: Just thought removing it will be less confusing, and since `roleBody` is an object created/populated on save, and since no error from the submit after doing so.

I've made the change now help check/review @ushahidlee

apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
```diff
@@ -136,6 +134,7 @@ export class RoleItemComponent implements OnInit {
         },
       });
     } else {
+      roleBody.name = this.role.name; // Use role name from API - We don't want to change name property along with display_name on edit/update
       this.rolesService.updateRole(this.role.id, this.form.value).subscribe({
```
ushahidlee: The change is correct, but you arent submitting the roleBody in the updateRole call.

ushahidlee (changes requested):
See comments.... super close!

apps/web-mzima-client/src/app/settings/roles/role-item/role-item.component.ts
Ifycode: You mean like this?
````
this.rolesService.updateRole(this.role.id, roleBody).subscribe({
````

Ifycode: Ok, I think I now understand the problem with my code it's not just there 🤭

Ifycode: Please check/review again @ushahidlee

ushahidlee (approved):
The last commit works perfectly and the extra cleanup is great too.

Ifycode:
Thank you @ushahidlee

tuxpiper:
Cherry-picked to development branch here: https://github.com/ushahidi/platform-client-mzima/commit/40f18171d417f9290392aee589da9b5f89a25107

PR #764, opened 2024-02-09
Displaying sub-categories in client

Angamanga:
This change make sub-categories (also called "child-categories") visible in the view in the following places:

1. Settings -> Categories
2. Settings > Surveys > Edit Survey > Edit Category
3. Settings > Surveys > Add Survey > Add Category-field
4. Add post with a category-field
5. Edit post with a category-field
6. Post-details view in data-view
7. Filter->categories

To test:
0. Prerequisite: at least one category with subcategories
1. Go to settings -> Categories
2.  - [ ] Each category with subcategories has an arrow, clicking on the arrow expands the list with subcategories
3. Go to settings -> Surveys
4. Add a new survey (or edit an existing one)
5. Add a category-field
6.  - [ ] The modal that opens when adding/editing a category-field should have subcategories visible
7. Add a new post to a survey that has categories with subcategories
8.  - [ ] The list of categories should have subcategories visible
9. Go to data-view and find the newly added post
10. Look at the details for the post
11. - [ ] The list of categories should have subcategories visible
12. Edit the same post
13. - [ ] The list of categories should have subcategories visible
14. Click on the filter-dropdown for categories
15.  - [ ] Each category with subcategories has an arrow, clicking on the arrow expands the list with subcategories


Ifycode (commented):
In my opinion, if this is just about display of categories and sub-categories on the UI - expanding and collapsing, and if they lead to the right page, then this PR is ready and approved.

#

> For how things should work after that, there are problems. I will add notes on a few of such problems on slack - incase they need to be treated differently or as separate issue.

#

Over to you @ushahidlee

ushahidlee (approved):
Looks good and does what it should. 

We have some interesting behaviour when selecting child categories that needs investigating, but it can be in a new ticket.

Issue #773, opened 2024-02-13
General Settings - Map


theobadha:
**Verify the map would work on the CI environment for this task.
In General Settings page:

- [x] Type a location in the Location field. Let's use the location "Nairobi"
- [x] Select the first option that pops up 
- [x] Verify the Latitude value changes to match selected option "-1.3026148499999999"
- [x] Verify the Longitude value changes to match selected option "36.82884201813725"

PR #780, opened 2024-02-15
E2e tests for General Settings Map

shakirandagire:
@AmTryingMyBest Could you help me review this PR that handles issue [773](https://github.com/ushahidi/platform-client-mzima/issues/773)


e2e-testing/cypress/locators/GeneralSettingsLocator.js
theobadha, line 1: - Append `Btn` and `Field` to the appropriate element locators. This would be for easy readability
- some attributes have a dash `-` as a separator while some have underscore `_` .Let's standardize this and use one

PR #783, opened 2024-02-15
[USH-1139] Data view: "There are no posts yet!" flicker always shows up

Ifycode:
**Fixes** "There are no posts yet!" flicker always showing up before content displays so that it does not appear/happen:
- On page load
- On page reload
- On Selecting filters:
     - Survey filters
     - Source filters
     - All other filters
- On sorting
- When you navigate to data view with the sidebar button/link
- On Clearing (all) filters
- By post card click - to switch from split mode to split view mode
- By clicking the mode switch buttons just above the sorting button -  to switch from split mode to split view mode (and vice versa)

**Plus Fix 🎁🥳:** 
- This Fix automatically takes care of the first part of [USH-1151](https://linear.app/ushahidi/issue/USH-1151/data-view-improve-user-experience-when-loading-more-content-in-split) - i.e. Load More button is now visible in split view


Ifycode:
Usually it's "small but mighty", but this one ended up being "mighty but small". 
What an issue ticket... Whew! 
Over to you @Angamanga @ushahidlee 
PR ready for review

apps/web-mzima-client/src/app/feed/feed.component.html
```diff
@@ -227,14 +227,13 @@ <h3 class="menu-title">{{ 'app.mark_as' | translate }}</h3>
...
             </app-post-preview>
           </ngx-masonry>
-          <app-spinner class="feed__loading" *ngIf="isLoading"></app-spinner>
           <mzima-client-button
             fill="clear"
             color="secondary"
             (buttonClick)="loadMore()"
             class="feed__load-more"
             *ngIf="
-              !isLoading &&
```
Angamanga: I think you can keep isLoading here?

apps/web-mzima-client/src/app/feed/feed.component.ts
```diff
@@ -314,6 +315,20 @@ export class FeedComponent extends MainViewComponent implements OnInit {
     });
   }
 
+  public postIsLoading(value: boolean) {
+    return (this.isLoading = value); // "There are no posts yet!" flicker is fixed for (most) places where the value here is set to true
```
Angamanga: I think you can just skip this function and use this.isLoading = value where needed directly instead.

apps/web-mzima-client/src/app/feed/feed.component.ts
```diff
@@ -314,6 +315,20 @@ export class FeedComponent extends MainViewComponent implements OnInit {
...
+  }
+
+  public postsCheck() {
+    const postsHaveLoaded = this.posts.length > 0;
+    if (postsHaveLoaded) this.postIsLoading(false); // post card/content area and LoadMore button benefits from this
+    const posts = {
+      atLeastOneExists: postsHaveLoaded,
+      stillLoading: this.isLoading, // tracks this.isLoading for post card area content display
+    };
```
Angamanga, line 324: If you use "this.isLoading" in the code, you don't need this stillLoading and then you can return postsHaveLoaded directly instead, to avoid complexity :)

Angamanga (changes requested):
Nice to figure this out, its really happening in many places! Just a few comments:
- You can use isLoading directly instead (like before) and refactor postsCheck to only return postsHaveLoaded. It decreases the complexity a bit I think?
- You can also remove the function postIsLoading and set this.isLoading directly where it is used 👍 . 
Then you would get the same result but with less code :) Otherwise, it is working fine and solves the annoying problem 🎉

apps/web-mzima-client/src/app/feed/feed.component.ts
Ifycode: Ok, this one is clear - I will change it. 

> PS: I just thought this will be more easy to read with the function name.

apps/web-mzima-client/src/app/feed/feed.component.ts
Ifycode: **1 of 2: stillLoading**

Ok, the thing is, the `stillLoading` is not necessarily the same as `isLoading`.

The first one is the `isLoading` - this only takes care of the loader that covers up the whole page.

Without the `stillLoading` (speaking based on the check `if (postsHaveLoaded) this.postIsLoading(false);`), the loader does not disappear even though the posts have loaded.

What `stillLoading` does is to work with the post length to dynamically determine when it's time for the loader to leave the screen and the posts content to become visible - so maybe the `stillLoading` is poor naming? But that's all it does.

`isLoading` does not work straight up like that without that small push from `stillLoading`.

#

**2 of 2: postsHaveLoaded**
Returning `postsHaveLoaded` also does not work does not fix the issue.

#

For now, handling this in that one method/function is the only & clean way I've been able to keep both the loader and the flicker in check.  Otherwise, one of these 2 misbehaves.

Angamanga: Ahh, yes, I understand now, it makes sense! Thanks for the explanation 🙏

apps/web-mzima-client/src/app/feed/feed.component.ts
Angamanga: Yes, its a matter of taste really :D It still works, my thinking was that we use this.isLoading in many places already so its good to use the same pattern.

apps/web-mzima-client/src/app/feed/feed.component.html
Ifycode: Resolved.

apps/web-mzima-client/src/app/feed/feed.component.ts
Ifycode: All right - resolved.

apps/web-mzima-client/src/app/feed/feed.component.ts
Ifycode: Ok, I'm yet to be able to do this differently, incase you have more suggestions.

Angamanga (approved):
All good and works fine 🎉

PR #784, opened 2024-02-16
USH-1040 - Add a text hint on date fields for post submission.

ushahidlee:
As in the ticket, added a little hint about default value to the date field. 

https://linear.app/ushahidi/issue/USH-1040/add-text-hint-to-date-fe


webong (approved):
Looks good to me!

PR #785, opened 2024-02-17
Ush 1020 - Add a title and some buttons to the date time picker on Mobile

ushahidlee:
https://linear.app/ushahidi/issue/USH-1020/add-select-time-to-post-creation-time-picker-fe


webong (approved):
capacitor upgrade always popping up 🥲 ... anyways it looks good to go.

PR #788, opened 2024-02-20
USH-1069 - Remove Search field from Collections screen

ushahidlee:
Removed the search field and tweaked a bit of the styling

https://linear.app/ushahidi/issue/USH-1069/remove-search-bar-from-collections-page


webong (approved):
Looks good to go!

PR #790, opened 2024-02-20
Fix (USH-927): Back button not working once you try to search for a deployment 

webong:
- Fix in-app back button during deployment search to fall back to choose deployment screen
- Fix hardware back button not closing app in android app


ushahidlee (approved):
Looks good, maybe remove the console.log lines though

PR #795, opened 2024-02-26
Ush 1123 - unstructured posts didn't return in the correct order

ushahidlee:
Issue: The ticket describes unstructured posts being in the wrong order implying a problem with sorting. However the actual issue as shown in the video is that unstructured posts are excluded entirely.

Solution: This ticket changes the default behaviour to include unstructure posts in results.


Angamanga (approved):
Looks good!

PR #807, opened 2024-03-05
add alert role where there's an alert to improve accessibility

JUNNY12:
- After searching  for class="alert" or class="alert error"  in the codebase of platform-client, I found it in two files survey-task.component.html and data-export.component.html
- I add role=alert in the html element in the two files
- The image is showing below after adding the roles to the html element 

edit1
edit2

-When pushing to my branch I encountered an error with the image below which made me add a xregexp.d.ts to the root folder to fix the error
Screenshot 2024-03-05 162049


Angamanga (approved):
Thank you @JUNNY12! You have followed the specifications in the issue and I will approve the PR because of that. However, I will not merge it because I realise this was an issue concerning how we used the class "alert" in the old codebase, and how we do it now is different. To learn more on when to use the role="alert", find info here: https://www.w3.org/WAI/ARIA/apg/patterns/alert/. Thank you for your contribution 👍

JUNNY12:
Thank you for approving my first Pull request @Angamanga ? If possible do you still want me to work on it following the guideline you sent

PR #814, opened 2024-03-05
fix Change placeholder text where all filter parameters are cleared #4735

Felysilver:
As per this issue [https://github.com/ushahidi/platform/issues/4735](url) I have fixed the placeholder text to be "Select a survey to show posts" even the issue discription suggested "select survey to show posts" 
select

### DIFF
--- a/apps/web-mzima-client/src/assets/locales/en.json
+++ b/apps/web-mzima-client/src/assets/locales/en.json
@@ -765,7 +765,7 @@
     "all_posts": "All Posts",
     "no_search_results": "Your search didn't match any posts",
     "no_search_results_location": "Your search didn't match any posts with location information",
-    "no_posts_yet": "There are no posts yet!",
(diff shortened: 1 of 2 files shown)


Felysilver:
@Angamanga, @tuxpiper I kindly ask if you can review this. Thanks!

Felysilver:
@Angamanga may have a look at this PR. Thanks!

Angamanga:
@Felysilver Thank you for your PR! Could you double-check so that string you changed are not used in other instances, for example, when there are actually no posts available?

Angamanga (changes requested):
Could you double-check so that string you changed are not used in other instances, for example, when there are actually no posts available?

Felysilver:
Thanks @Angamanga, I got your point, I am working on it!

Angamanga:
This is fixed

PR #815, opened 2024-03-05
fix:Replacing-no-posts-yet-text

jatinder14:
Fixes: #4735


jatinder14:
Fixes https://github.com/ushahidi/platform/issues/4735#event-12029371102

jatinder14:
Outreachy Task Submission

Angamanga (changes requested):
Thank you for your PR @jatinder14! Could you double-check so that string you changed are not used in other instances, for example, when there are actually no posts available? I suspect it is used in more cases, for example when there are no posts in a deployment, and then the text is actually correct. So there needs to be some logic added for when to show what text.

jatinder14:
@Angamanga can you please check it again ?

apps/web-mzima-client/src/app/feed/feed.component.html
```diff
@@ -272,5 +272,10 @@ <h3 class="menu-title">{{ 'app.mark_as' | translate }}</h3>
 </ng-template>
 
 <ng-template #postsEmpty>
-  <span class="posts-empty">{{ 'post.no_posts_yet' | translate }}</span>
+  <span class="posts-empty" *ngIf="filters.form[0] === 'none'; else noPosts">{{
+    'post.no_posts_yet' | translate
+  }}</span>
+  <ng-template #noPosts>
+    <span class="posts-empty">There are no posts in the survey!</span>
```
Angamanga: Add the string to the en.json-file instead and map it in a similar way as is already done above, on row 276

Angamanga (changes requested):
Great start! See my comment in the code for directions 👍

apps/web-mzima-client/src/app/feed/feed.component.html
jatinder14: alright @Angamanga But it should be done for english only or for all the available languages?

jatinder14:
@Angamanga  can you please check it again ?

jatinder14:
@Angamanga

Angamanga (approved):
Great work! Approved 👍

Angamanga:

jatinder14:
please merge it @Angamanga

jatinder14:
Hi @Angamanga @tuxpiper

jatinder14:
@Angamanga

Angamanga:
Fixed

PR #827, opened 2024-03-07
USH-1125: Show location fields as available to import CSV data into

Angamanga:
@ushahidlee Here is my wip code for reference. 
https://linear.app/ushahidi/issue/USH-1125/import-csv-doenst-show-the-location-field-as-option-of-mapping


tuxpiper (approved):
Looking good to me

PR #830, opened 2024-03-07
[Outreachy Task Submission] Buttons Must have discernible text for screen readers 

MundiaNderi:
# Description
Some of the buttons like the password hide button on the login pages and the share button on the feed pages have no inner text so adding aria-label will help in accessibility ie the filter and the share-modal component.

For the mzima button which is a component used across the site, the earlier implementation does not allow for aria-labels and adding one throws an error. So I added  a span with text to help in screen reader accessibility, then I added a css class that hides the span from the naked eye ie in the registration, collection and search form components.

## Fixes 
This PR fixes the issue: https://github.com/ushahidi/platform/issues/4761 and https://github.com/ushahidi/platform/issues/4763

## Testing
To test: 
- Install [axeDevTools ](https://chromewebstore.google.com/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd)on your chrome browser and scan the pages where the file have been changed - no button must have discernible text error
- remove the aria-label or the span element and scan the page. You will get a 'Button must have discernible text error' You can also highlight to see which icon/button is problematic.


MundiaNderi:
@Angamanga this PR is ready for review.

Angamanga:
Hi @MundiaNderi! Thank you for your PR! Usually, we aim to solve one issue per PR, it looks like this PR solves 3 issues? Could you try separating the things you have fixed into three separate PRs? OR, if they are very related, add a description to this PR describing all changes made as well as a checklist on how to test the changes? Thank you :)

MundiaNderi:
@Angamanga all the issues were related, in that the buttons and icons were not screen reader supported so I thought that having many PRs would be redundant, but going forward I'll keep it to one issue per PR. I have made the changes and I have added instructions on how to test the changes.

Angamanga:
Ok, thanks for the clarification @MundiaNderi 🙏

Angamanga (changes requested):
What error do you get when trying to add the aria-label to the mzima-button? It is possible to add an aria-label, to the button, but the syntax may be a little bit different, look at how it is solved here: 
https://github.com/ushahidi/platform-client-mzima/blob/development/apps/web-mzima-client/src/app/shared/components/filter-control/filter-control.component.html#L147

MundiaNderi:
Screenshot 2024-03-18 at 22 45 29
 @Angamanga  this is the error. I'll explore the posted solution and revert.

MundiaNderi:
@Angamanga I went through your proposed solution and even though it didn't work for me, I found the solution.
I found the button component which generically sets the aria-label dynamically based on the ariaLabel property and ended up using the ariaLabel property for the mzima button. [attr.aria-label]="ariaLabel" . I also deleted the css classes that I had used to set/hide span class.

This PR is now ready for review. 

Screenshot 2024-03-20 at 09 30 53

package-lock.json
Angamanga, line 1: Remove this file from the PR 👍

Angamanga (changes requested):
Almost there, just remove the package-lock file from the PR 👍

MundiaNderi:
@Angamanga happy to report that this is fixed.

Angamanga (approved):
Excellent! Approved :) 👍

PR #832, opened 2024-03-07
fix: disable viewing and updating collections without correct permissions

mannyAndem:
This pull request fixes issue [4739](https://github.com/orgs/ushahidi/projects/18/views/1?pane=issue&itemId=55115848) which focused on disabling access to view and modify collections without having the correct permissions.


Angamanga:
Thank you for your PR @mannyAndem  🙏. Could you please add a short description and a checklist for testing the fix as well and then we'll review it 👍

mannyAndem:
@Angamanga sorry for replying late, the internet downtime hit real bad. But I'll do this pronto :)

mannyAndem:
This pull request fixed the issue of allowing only users with the correct privileges to view and update collections. Only users with "read" permissions can view collections while only those with "update" permissions can add a post to a collection. Posts with visibility status of "everyone" will be seen regardless of the permission level of the current user (even if the user isn't logged in at all).

To test the fix:

- [ ] As a user with admin privileges, create a collection and set its visibility to any level except everyone.
- [ ] Log out.
- [ ] You shouldn't be able to view the newly created collection.
- [ ] Collection should be visible to everyone with read privilege.

Please let me know if there's anything that requires tweaking @Angamanga :)

mannyAndem:
I rewrote the permissions logic to be encapsulated in a ```checkReadPrivileges``` function which checks if user fulfils any of the conditions necessary to view a specific collection. @Angamanga :)

Angamanga (changes requested):
Thank you for your contribution @mannyAndem! It seems like I still can add posts to collections when not having the correct priviliges, if I login as a "member", I can still add posts to collections, which I should not be able to. Could you please take a look?

mannyAndem:
Hmm, that's odd, I might have overlooked something. Will check right now!

mannyAndem:
@Angamanga, I think I've got it. Hid the add post to collection button when user does not have any of the "Manage Posts" or  "Manage Collections and Saved Searches" permissions as per the user documentation. Could you take a look now? :)

apps/web-mzima-client/src/app/shared/components/collections/collections.component.ts
```diff
@@ -149,15 +151,37 @@ export class CollectionsComponent extends BaseComponent implements OnInit {
...
             ...item,
             my_collection: isOwner,
-            visible: this.isManageCollections || !(item.featured && !isOwner),
+            visible: checkReadPrivilege(),
           };
         });
         this.collectionList = this.collectionList.filter((collection) => collection.visible);
         this.isLoading = false;
+        console.log(this);
       },
```
Angamanga: Nearly there, just remove this console log :)

Angamanga (changes requested):
Good job, just a tiny change to remove a forgotten console.log :)

mannyAndem:
Done @Angamanga :)

Angamanga (approved):
Thank you! Approved!

mannyAndem:
Arigato @Angamanga :)

PR #836, opened 2024-03-08
[Outreachy Task Submission] Accessibility Fix For Keyboard Navigation in Help & Support

### LINKED ISSUE #4774: Accessibility Issue with Keyboard navigation on 'Help & Support' screen
# Intro
This is an accessibility issue I found with the 'Help & Support' screen.

## Issue
Keyboard navigation using the 'tab' key does not work at all for navigating through the options on the screen. This is a violation of WCAG Success Criterion 2.1.1 (Level A) - keyboard. [See](https://www.w3.org/WAI/WCAG21/Understanding/keyboard.html)

## Description
Success Criterion 2.1.1 (Level A) - keyboard requires that all functionality of the content is operable through a keyboard interface without requiring specific timings for individual keystrokes, except where the underlying function requires input that depends on the path of the user's movement and not just the endpoints.

## Impact
People who are blind will be unable to get help and support from the service. People with low vision who find it difficult to track pointers will encounter difficulty, and people with hand tremors who cannot use a computer mouse will also experience navigation difficulties. 

## Fix
The list options would need to be configured to allow for keyboard tab navigation.

Chiemezuo:
Fixes [#4774](https://github.com/ushahidi/platform/issues/4774)

I added `(keydown.enter)` attributes to the `matRipple` divs with their appropriate mappings.
I also tested with my screen reader app and it worked fine,

## How to Test the PR.
1. Open your browser and log into your running Ushahidi deployment (from localhost).
2. Navigate to the 'Help & Support' section.
3. When the modal opens, click on any white/blank space on the modal to put it in focus.
4. The idea is to traverse each menu item, so keep clicking the `tab` on your keyboard till you can do this.
5. Notice that the menu items cannot be navigated to or receive focus.
6. Switch off the deployment from your terminal.
7. Checkout to this PR branch.
8. Re-run your deployment from the Localhost.
9. Repeat steps 1-4.
10. Notice you can now navigate through the menu items with the `tab` button.
11. Also try pressing the `enter` key while any menu item is in focus.
12. Notice that navigation with `tab` and `enter` keys works this time around.

### DIFF
--- a/apps/web-mzima-client/src/app/shared/components/support-modal/support-modal.component.html
+++ b/apps/web-mzima-client/src/app/shared/components/support-modal/support-modal.component.html
@@ -14,7 +14,7 @@
 <strong mat-dialog-title>{{ 'nav.help_support' | translate }}</strong>
 
 <div mat-dialog-content>
-  <div
+  <a
     matRipple
     class="menu-item"
     (click)="item?.action()"
@@ -24,12 +24,12 @@
   >
     <h4 class="menu-item__title">{{ item.title }}</h4>
     <p *ngIf="item.description?.length">{{ item.description }}</p>
-  </div>
+  </a>
   <!-- <div matRipple class="menu-item" *ngIf="isLoggedIn">
     <h4 class="menu-item__title">{{ 'app.intercom.intercom' | translate }}</h4>
     <p class="intercom_custom_launcher">{{ 'app.intercom.description' | translate }}</p>
   </div> -->
-  <div
+  <a
     matRipple
     class="menu-item"
     (click)="openUrl('https://www.ushahidi.com/terms-of-service/')"
@@ -37,8 +37,8 @@
     tabindex="0"
   >
     <h4 class="menu-item__title">{{ 'app.terms_and_conditions' | translate }}</h4>
-  </div>
-  <div
+  </a>
+  <a
     matRipple
     class="menu-item"
     (click)="openUrl('https://www.ushahidi.com/privacy-policy/')"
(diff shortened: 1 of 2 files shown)


Angamanga:
Thank you for your PR @Chiemezuo  🙏. Could you please add a short description and a checklist for testing the fix as well and then we'll review it 👍

Chiemezuo:
I've updated the comment with a guide on how to test the changes. @Angamanga

Angamanga (changes requested):
I think this violation depends on the fact that we are not using the correct html. If we would use correct links instead of divs, tabbing would be fine. Would you mind trying to exchange the html to correct syntax and use CSS for the styling instead?

Chiemezuo:
@Angamanga I changed the elements from `</div>` to `</a>` tags, and the tab issue still persisted up until I set the `tabindex` property. I also had to modify the accompanying CSS as well. I'll raise a PR so you can look at it, but it took more code to still end up needing to set the `tabindex` property at the end of the day.

Here's the CSS I had to add to the `.menu-item` class:
```css
 /* Additional styles for anchor tag */
  text-decoration: none; /* Remove default underline */
  display: block; /* Anchor tags are inline by default, change to inline-block for padding and border-radius to take effect */
```
If we are to opt for less bulky code, leaving things as is and only setting a `tabindex` works better.
If we are to opt for more semantic code, however, your suggestion is much better in the short and long term.

Chiemezuo:
@Angamanga I've raised a new PR with your suggestions.

PR #859, opened 2024-03-11
[Outreachy Task Submission] Fix - Forms making unnecessary API requests to back-end

RashJrEdmund:
updated the applyFilters() and loadData() methods in the search-form component.ts and collections.component.ts files respectively to prevent making any API requests if no search term of input field value is present.

Fixes issue https://github.com/ushahidi/platform/issues/4788

[Screencast from 03-11-2024 11:05:43 AM.webm](https://github.com/ushahidi/platform-client-mzima/assets/116172031/fa13003c-630b-46ac-b9cb-1567b7e9be08)

### DIFF
--- a/apps/web-mzima-client/src/app/shared/components/collections/collections.component.ts
+++ b/apps/web-mzima-client/src/app/shared/components/collections/collections.component.ts
@@ -133,6 +133,8 @@ export class CollectionsComponent extends BaseComponent implements OnInit {
   }
 
   loadData(query = '') {
+    if (!query.trim()) return;
+
     this.isLoading = true;
     let params: any = new Map();
     params = {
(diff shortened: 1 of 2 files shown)


Angamanga (commented):
Hi @RashJrEdmund Technically, this works, but I am a bit unsure about the behaviour. Someone writes a query, looks at the results, and then they remove the query, wouldn't they expect to see all the posts, not just the ones that matched the previous text? Do you have any thoughts around that?

RashJrEdmund:
hi @Angamanga , thanks for reviewing, but what I meant was that there is no need to make an API request if the input field is empty.

The thing is when the form is submitted with an empty input field, it still tries to make a request to the backend, and there is no need making an empty request to the API. This will help reduce the number of hits to the backend.

Angamanga (approved):
Thank you for your contribution! Approved!

Angamanga:
@Angamanga To consider after filter-changes are done!

PR #871, opened 2024-03-12
Support/mobile issues ush 991 1099

webong:
- Resolve onboarding bug with loading backend_url from angular environmental variable.
- Remove feature input field from create/edit collection.

Fixes USH-991 USH-1099


ushahidlee (approved):
Looks good and works as described :+1:

PR #874, opened 2024-03-12
Ush 1203

ushahidlee:
Fixed a stupid order of execution bug which resulted in erroneous assignment of required fields. 

Fixes: https://linear.app/ushahidi/issue/USH-1203/mapping-required-fields-in-survey-to-fields-in-csv-file-fails


webong (approved):
Code looks good to me.

PR #884, opened 2024-03-14
[Outreachy Testing Task] Adding Cypress Testing Tool and integrating it with axe-core

MundiaNderi:
# Description
For this Outreachy testing task, I used Cypress integrated with axe-core. Cypress is a widely used E2E testing framework and when powered with axe-core, generates accessibility reports.  I chose Cypress because it has  a great testing reputation and axe-core  because of its accessibility testing reputation.

In my testing scripts, I used Cypress to 'visit' url pages locally and axe-core to check for accessibility.  For this, I added the following files:
- a cypress/support/e2e.ts file where I import cypress.axe which ensures that cypress-axe is imported every time our tests run.
- a ts.config.json without which Cypress throws an error.
-  After you run the first e2e test, Cypress adds an e2e/spec.cy.ts folder here you can add more tests or use the same file to test various pages. Make sure to change the URL address of the site/page you're visiting .

## Evaluation
- Why did you select this tool ? In the past, I had used Cypress as an E2E testing feature, and axe-core as a chrome extension to test accessibility. When I saw that you could combine the two to test accessibility, I found that very neat because Cypress is a well known robust and reliable web testing framework.

- Does the Output make sense to you ? Yes the output make sense to me - in the sense that it asserts whether or not there are errors, how many errors and you can look at the stack trace to see where the errors are occurring. 

- Was it easy or really hard to implement ? It was easy because I had prior Cypress experience but if this was the first Time I reckon it would be a little hectic. [CircleCi ](https://circleci.com/blog/cypress-accessibility-testing/)has a great tutorial on how to go about it.

- Do you trust the output or do you think it misses something ? I trust the output.  It shows me where the accessibility errors are, and what is missing from the screenshot below.

- Any other comments that you'd like to share: Cypress is a great tool when paired with axe-core. I find it pretty neat because one can integrate accessibility testing as they conduct their E2E test, instead of testing for accessibility at a later point in time.

Screenshot 2024-03-14 at 10 11 27
Screenshot 2024-03-14 at 10 15 18

## Fixes 

Fixes issue https://github.com/ushahidi/platform/issues/4808


MundiaNderi:
@Angamanga this PR is ready for review.

PR #892, opened 2024-03-15
USH-1212 Survey Fields are duplicated in the mapping section

ushahidlee:
This PR fixes two things:

1) Removes a leftover comment in the template that was showing in the frontend.

2) Fixes an issue where selecting multiple surveys could result in the title and description fields being duplicated.

https://linear.app/ushahidi/issue/USH-1212/survey-fields-are-duplicated-in-the-mapping-section


Angamanga (approved):
🌮 🎉 🌮

PR #895, opened 2024-03-16
[Outreachy Testing Task] Add Automation Tool For Accessibility Test

JUNNY12:
Tool Used : Pa11y

### Description:
Added Pa11y for accessibility testing of the codebase. Pa11y is a popular accessibility testing tool that allows  to automate accessibility testing and ensure application is accessible to all users.

### Why Pa11y
I selected Pa11y because of its ease of use and comprehensive testing capabilities

### How do you like it
I found Pa11y easy to use, with a simple command-line interface and clear documentation and also very easy to implement with the existing code base

### Does the Output make sense
The output provided by Pa11y makes sense to me, as it highlights specific accessibility issues by running the command line to test individual URL or has a whole. Below is an image:

pa11y

### Was it easy or really hard to implement?

It was very straightforward to implement without any issue

###  Do you “trust” the output or do you think it miss something?
I trust the output of Pa11y, as it is based on established accessibility standards and guidelines. You can also set the specific guideline to follow in pa11y.json

### Any other comments or insights you would like to share
I noticed some inconsistency when running the command line, although I don't know maybe its an issue with my setup, it doesn't brings the same accessibility  report when ran multiple times on same URL, although I noticed that the reports are valid


JUNNY12:
@Angamanga Here is my Submission for the tooling-task. Please kindly review

PR #901, opened 2024-03-16
[USH-1184] Change hint text in Post Search page

MMwandigha:
Adjusted text hint on Posts Search Page to show 'Search for Posts' instead of 'Search for site or Enter URL'


Angamanga (approved):
Works fine 👍

PR #910, opened 2024-03-17
Fix font size issue

mannyAndem:
This pr fixes the font-size issue outlined in issue [4790](https://github.com/ushahidi/platform/issues/4790). I created sass variables to handle the font-sizes in the helpers.scss file. So, browser font-size settings as well as font-size styles inputted by external stylesheets (such as stylus) would be reflected in the site. This makes the site way more accessible to users with low vision who may not be able to read the smaller sized text clearly.

I also went a step further to make the site usable at 200% font-size settings as outlined by the WCAG for AA standard. It may not be perfect but it's a start, screenshot at 200% below :)

2024-03-17 14_43_23-Demo

### DIFF
--- a/apps/web-mzima-client/src/app/activity/activity-timeline/activity-timeline.component.scss
+++ b/apps/web-mzima-client/src/app/activity/activity-timeline/activity-timeline.component.scss
@@ -1,5 +1,4 @@
 @import 'helpers';
-@import 'typography';
 
 .chart-cover {
   overflow-x: auto;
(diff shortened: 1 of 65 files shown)


Angamanga (approved):
Thank you for your contribution! Approving but will not merge just yet since work is ongoing for css-structure, naming and I would also like to discuss this with our product-team. But its approved since a nice fix and technically correct 🙏

Angamanga:
@Angamanga For best-practices work.

mannyAndem:
Thank you @Angamanga :)

PR #917, opened 2024-03-19
Feature/cancel

kivuvarosekivuvan:


Aquilaafuadajo:
Hello @kivuvarosekivuvan so after debugging this, here is what noticed 👇🏽 
1. on line https://github.com/ushahidi/platform-client-mzima/blob/90528f3d8dc332d5d9e3712121be4179cc99ac05/apps/web-mzima-client/src/app/settings/general/general.component.ts#L31-L33
you don't need to declare initialFormData and form. The data is already being stored in this.generalForm
2. on line https://github.com/ushahidi/platform-client-mzima/blob/90528f3d8dc332d5d9e3712121be4179cc99ac05/apps/web-mzima-client/src/app/settings/general/general.component.ts#L187 this.form.value will be undefined so it'll cause the objectHelpers function to throw error.

To fix the issue you're having, you can store a copy of generalForm in initialFormData when the component mounts then use that to compare in your confirmModal function

PR #921, opened 2024-03-19
Removing call to v3-endpoint for post-details

Angamanga:
Removing an extra call to V3 to get form-info for displaying form-name when displaying a post.
To test:
1. Go to data-view
2. Click on a post
3. The post should display correctly with all form-info present


ushahidlee (approved):
Looks good and does what its supposed to :+1:

PR #923, opened 2024-03-19
 Resolving Accessibility Issue: Inadequate Color Contrast

Mekebib-Ergudo:
To address the accessibility issue, the text color was changed to a color that provides sufficient contrast against the background color. This adjustment significantly increased the contrast ratio from 1.45, which falls below the WCAG standard, to 19.05, which is highly recommended in the WCAG standard.

Screenshot 2024-03-19 at 10 15 18 AM

### DIFF
--- a/apps/web-mzima-client/src/app/settings/data-export/data-export.component.html
+++ b/apps/web-mzima-client/src/app/settings/data-export/data-export.component.html
@@ -13,7 +13,8 @@
           <p>{{ 'data_export.hxl_apikey_alert_1' | translate }}</p>
           <p>
             {{ 'data_export.hxl_apikey_alert_2' | translate }}
-            <a [routerLink]="['/settings/user-settings']" class="link-blue">
+            <!--Inadequate Color Contrast for Accessibility -->
+            <a [routerLink]="['/settings/user-settings']" class="colorLink">
               {{ 'data_export.hxl_configure' | translate }}
             </a>
             {{ 'data_export.hxl_apikey_alert_3' | translate }}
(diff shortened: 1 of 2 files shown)


apps/web-mzima-client/src/app/settings/data-export/data-export.component.html
```diff
@@ -13,7 +13,8 @@ <h1>{{ (hxlEnabled ? 'data_export.title_hxl' : 'data_export.title') | translate
           <p>{{ 'data_export.hxl_apikey_alert_1' | translate }}</p>
           <p>
             {{ 'data_export.hxl_apikey_alert_2' | translate }}
-            <a [routerLink]="['/settings/user-settings']" class="link-blue">
+            <!--Inadequate Color Contrast for Accessibility -->
+            <a [routerLink]="['/settings/user-settings']" class="colorLink">
               {{ 'data_export.hxl_configure' | translate }}
```
Angamanga, line 18: A question here, what happens if your entirely remove the class?

Angamanga (changes requested):
Two questions for this change. Is the class link-blue used somewhere else? Would it make sense to change the css for that class instead of adding a new one?

apps/web-mzima-client/src/app/settings/data-export/data-export.component.html
Mekebib-Ergudo: Hello there  Yes, the class "link-blue" is used elsewhere, and when I tried to add a new class name, the original class property was overriding the new class property. Therefore, I removed the original class and set the text color to black for the new class. and removing original the class is not affect any othere  properties.

Angamanga (approved):
Thank you for the explanation 👍 Approved!

PR #924, opened 2024-03-20
USH 1204 - Errors in console after CSV Import

ushahidlee:
This PR fixes some issues following a CSV import, specifically allowing the user to be notified that the job was successful or not. Previously this functionality was incomplete, meaning the status page would endlessly loop, the user would never be notified of the job's success or failure and the console would fill up with errors.

https://linear.app/ushahidi/issue/USH-1204/messy-console-with-error-code-500-when-attempting-import


apps/web-mzima-client/src/app/settings/data-import/import-results/import-results.component.ts
```diff
@@ -25,6 +25,7 @@ export class ImportResultsComponent implements OnInit {
 
   ngOnInit(): void {
     const jobId = this.route.snapshot.queryParamMap.get('job')?.split(',');
+    console.log(jobId);
     if (jobId) {
```
Angamanga: Remove the console.log 🙃

apps/web-mzima-client/src/app/core/services/polling.service.ts
```diff
@@ -65,14 +65,14 @@ export class PollingService implements OnDestroy {
...
-        result.forEach((job: ExportJobInterface) => {
-          if (job.status === 'SUCCESS') {
-            this.notificationService.showError('JOB SUCCESS SUCCESS');
-            this.importFinished.next(job);
-          } else if (job.status === 'FAILED') {
-            this.notificationService.showError('JOB FAILED');
+        result.forEach((job: any) => {
+          if (job.result.status === 'SUCCESS') {
+            this.showNotification('success');
+            this.importFinished.next(job.result);
```
Angamanga, line 70: It works, but the text in the notification says "Upload complete. You should now see your tagged data in your HDX account.", I think it is the title that is hardcoded on row 95. Maybe we need to send that from the start instead in order to be able to reuse the function?

Angamanga (changes requested):
It works but a change is needed in the showNotification-function in order to show the correct string in the notification-bar.

Issue #725, opened 2023-12-12
Survey Configuration - Hide exact Time information


theobadha:
Posts sent in have a timestamp of when they were submitted shown on the post. By default the option to hide the exact time information is disabled. In this test, we will enable hiding the exact time information and verify that exact time information is hidden in posts

Test Steps
- [x]  As an admin, while logged in, open a survey in edit mode Settings -> Surveys -> Click on a Survey
- [x]  Click on Configure Tab
- [x]  Hide exact time information is disabled by default
- [x]  Click toggle switch to enable Hide exact time information
- [x]  Save
- [x]  Log out as admin
- [x]  Submit a post to the survey above
- [x]  Verify that time information on the post is an ambiguous time from when the post is submitted

ccokeke0023:
Hey @AmTryingMyBest ,

I would like to  work on this issue.
Could you assign me the ticket so I can take it up?

Thanks

PR #931, opened 2024-03-21
E2e tests for Survey Configuration - Hide exact Time information

shakirandagire:
@AmTryingMyBest Could you help me review PR for [issue](https://github.com/ushahidi/platform-client-mzima/issues/725)


e2e-testing/cypress/functions/SurveyConfigurationFunctions.js
```diff
@@ -52,7 +56,9 @@ class SurveyConfigurationFunctions {
   }
 
   type_post_title(title) {
-    cy.get(SurveyConfigurationLocators.postTitleField).type(title, { force: true });
+    cy.get(SurveyConfigurationLocators.postTitleField)
+      .should('be.visible')
```
theobadha, line 60: this `should('be.visible')` assertion is what we should use in place of the explicit `cy.wait` for elements to be interacted with. Here I replaced `cy.wait(3000)`
it would be a worthy effort to go through the tests and remove the explicit waits. They are not good practice in tests.

e2e-testing/cypress/functions/SurveyConfigurationFunctions.js
```diff
@@ -91,6 +97,29 @@ class SurveyConfigurationFunctions {
     cy.get('.post-info__username').contains('Anonymous').should('be.visible');
   }
 
+  check_for_time_post_was_added() {
```
theobadha, line 100: How I checked for this is:
 - where a post has time hidden, the admin would be able to see the exact time a post was sent in, while a non-admin or user without the right privileges will see an inaccurate time
 - the check here then is log in as admin, see the time in the most recently sent in post, and log out and verify that the time seen while logged out is not the same.

e2e-testing/cypress/locators/SurveyConfigurationLocators.js
```diff
@@ -14,6 +14,7 @@ const SurveyConfigurationLocators = {
   postPreview: '[data-qa="post-preview"]',
   postItem: '[data-qa="post-item"]',
   clearBtn: '[data-qa="clear"]',
+  postDate: '[data-qa="date"]',
```
theobadha: it may be a good idea to rename the `data-qa="date"` to something better relatable like `"post-date"`

e2e-testing/cypress/functions/SurveyConfigurationFunctions.js
shakirandagire: Thanks for the correction. Will follow this from now onwards.

e2e-testing/cypress/functions/SurveyConfigurationFunctions.js
shakirandagire: @AmTryingMyBest this clarification actually makes sense to implement this way.

e2e-testing/cypress/locators/SurveyConfigurationLocators.js
shakirandagire: Thanks for the recommendation on it.

shakirandagire (commented):
@AmTryingMyBest The changes look good to me. Let me implement the feedback given.

e2e-testing/cypress/functions/SurveyConfigurationFunctions.js
theobadha: @shakirandagire follow this for hiding location test too

theobadha (approved):

PR #932, opened 2024-03-21
[Outreachy Task Submission] Fixed "All time" select range value from dropdown on the activity page

crocmons:
Resolves this issue [#4857](https://github.com/ushahidi/platform/issues/4857)
Fixes an accessibility issue where selecting "All time" from the dropdown menu in the activity page did not appear as a dropdown selected value. The issue was due to the use of `null` as the value for "All time" in the `ranges` array. **This PR replaces the `null` value with an empty string `" "` to ensure clear all time feedback to users. Also I saw that there is a typo of `displayNane` so I fixed this typo as well and wrote `displayName` for every file inside the activity component.**

**Changes Made**
- Updated the `ranges` array in `activity-page.component.ts` to use `" "` instead of `null` for the "All time" option.

**Demo**
 
[screencast-127.0.0.1_4200-2024.03.21-14_16_32.webm](https://github.com/ushahidi/platform-client-mzima/assets/106610322/3f1d9dea-574f-4ce3-b301-dc065bb5ce61)

**Accessibility Considerations**
This change addresses an accessibility issue by providing clearer feedback to users

**Testing**
- Manually tested the fix by selecting "All time" from the dropdown menu in the activity page.
- Verified that "All time" now appears as a dropdown selected value.

@Angamanga this pr is ready for review

### DIFF
--- a/apps/web-mzima-client/src/app/activity/activity-timeline/activity-timeline.component.ts
+++ b/apps/web-mzima-client/src/app/activity/activity-timeline/activity-timeline.component.ts
@@ -34,10 +34,10 @@ export class ActivityTimelineComponent implements OnInit {
   public cumulativeTotal = true;
   public dateRange: ManipulateType;
   public filters = [
-    { displayNane: 'Show all posts', value: '' },
-    { displayNane: 'Categories', value: 'tags' },
-    { displayNane: 'Survey', value: 'form' },
-    { displayNane: 'Status', value: 'status' },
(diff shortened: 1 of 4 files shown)


Angamanga (approved):
Wow @crocmons, such a great catch! Approved!

Angamanga:

crocmons:
> Wow @crocmons, such a great catch! Approved!

thank you @Angamanga

PR #935, opened 2024-03-22
fix: make search input accessible

mannyAndem:
This pr fixes issue [4858](https://github.com/ushahidi/platform/issues/4858) which talked about the search input being inaccessible to users relying on screen readers. This pr fixes that issue by adding an '''aria-label''' attribute with the same content as the placeholder text users with normal vision would see. I also made the button just beside it (the one with the search icon) invisible to ATs as it did not perform any specific purpose other than visual aesthetics. 

To test the fix:

- Access the site with a screen reader.
- Focus on the search input.
- Should be described with the same content contained in its placeholder text.

### DIFF
--- a/apps/web-mzima-client/src/app/app.component.html
+++ b/apps/web-mzima-client/src/app/app.component.html
@@ -15,7 +15,7 @@
       [languages]="languages"
       [selectedLanguage]="selectedLanguage$ | async"
     ></app-toolbar>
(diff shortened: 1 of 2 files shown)


Angamanga (approved):
Thank you for your contribution! Approved :)

Angamanga:

mannyAndem:
Thanks @Angamanga !

PR #936, opened 2024-03-22
add skip to main content link

mannyAndem:
This pr fixes issue [#4860](https://github.com/ushahidi/platform/issues/4860) which talked about users not being able to skip navigation panel. This fix adds a "Skip to main content" button that only becomes visible when tabbed into. Clicking it skips the navigation and search and gets to the main content of the page.

To test:

- Using the tab key, start navigating through the page.
- Just after the "Add new post" button and before the navigation, a "skip to main content" should appear.
- Activating it (by pressing space or enter key depending on browser settings) should skip to the main content of the page

### DIFF
--- a/apps/web-mzima-client/src/app/shared/components/main/sidebar/sidebar.component.scss
+++ b/apps/web-mzima-client/src/app/shared/components/main/sidebar/sidebar.component.scss
@@ -15,25 +15,6 @@
   -webkit-overflow-scrolling: touch;
   padding: 24px var(--side-offset) 20px;
 
-  .visibly-hidden {
-    position: absolute;
-    width: 1px;
-    height: 1px;
-    padding: 0;
-    margin: -1px;
-    overflow: hidden;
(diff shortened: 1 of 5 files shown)


apps/web-mzima-client/src/app/shared/components/main/sidebar/sidebar.component.scss
```diff
@@ -15,6 +15,25 @@
   -webkit-overflow-scrolling: touch;
   padding: 24px var(--side-offset) 20px;
 
+  .visibly-hidden {
```
Angamanga: Could you move this to styles.scss instead? I think its useful in more places than just here :)

Angamanga (changes requested):
Very nice addition :) Just a small change-request, see my comment in the code!

mannyAndem:
Done @Angamanga :)

Angamanga (approved):
Thank you! Approved!

mannyAndem:
Thank you!

PR #953, opened 2024-03-26
[Outreachy Task Submission]: Fix general settings bug

mannyAndem:
This pr fixes issue [4876](https://github.com/ushahidi/platform/issues/4876) which referenced a bug that occurs in the general settings menu. Apparently, it was caused by passing slightly incorrect data to the setConfigurations method. While it expects an object containing the updated site config, when setting changes are saved, it gets passed an object with a result property holding the actual data it requires.

## Steps to test fix

- Navigate to the general settings menu.
- Change a setting.
- Save the changes.
- Navigate away from the settings menu (to map view for example).
- Navigate back to the general settings menu.
- Changes should be reflected.

### DIFF
--- a/apps/web-mzima-client/src/app/settings/general/general.component.ts
+++ b/apps/web-mzima-client/src/app/settings/general/general.component.ts
@@ -57,6 +57,7 @@ export class GeneralComponent implements OnInit {
 
   ngOnInit(): void {
     this.siteConfig = this.sessionService.getSiteConfigurations();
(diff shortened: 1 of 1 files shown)


tuxpiper (approved):
Thank you for your contribution! This has been fixed since, but the contribution counts regardless 👍  Cheers!

mannyAndem:
Wasn't aware it had been fixed already @tuxpiper 😅, but thank you!

PR #958, opened 2024-03-26
Show collections in mobile app

Angamanga:
This PR shows collections to logged in users in the mobile app. To test
As a logged in user:
1. Go to "profile"
2. Tap on Collections
3. Tap on a Collection
- You should now be directed to the Map-view with only posts from that collection visible in the map/feed-list. 
- Above the feed-list there is a back-arrow, number of posts in the collection and collection-name.
- When tapping on the back-arrow, filters are reset and you see all posts again


ushahidlee (approved):
Code is great and the functionality works exactly as described. There is a slight issue in styling though where the back button is very close to the name of the collection selected.

unnamed

PR #977, opened 2024-03-28
E2e enhance collections

theobadha:


theobadha:
@shakirandagire lookie here. Two different tests were being run here. 
1. Create a collection
2. Add a post to a collection

I have separated the two tests. 
Create a collection runs fine, but adding a post to a collection is a bit more involving to get right. I've skipped that test in the test run, and created a task for completing the test.

PR #990, opened 2024-03-30
[Outreachy Task Submission] Prevent Weak Passwords from Creating Accounts

### LINKED ISSUE #4908: [Outreachy Task Submission] Weak Passwords Creating Accounts
# Introduction
In the current state of the Ushahidi platform, when trying to register an account, users are prompted when their passwords are too weak. However, once they hit the minimum character requirements, weak passwords can still create accounts.

## How to Reproduce this behavior
1. Open the Ushahidi platform (either via a deployment or Localhost)
2. Click on `Login & Signup`.
3. Switch to the Signup view and fill in the form (except the password form).
4. For the password, fill in something that only meets the minimum length constraints but ignores other constraints (for example "adminooo"). Note: you can copy and paste the example.
5. You will notice that the password hint suggests that your password is very weak.
6. Agree to the terms and conditions.
7. Create an account.
8. Account creation works.

## My proposal
When the password is very weak, users should not be allowed to create accounts so as not to undermine their security.

Chiemezuo:
# Introduction
This fixes [#4908](https://github.com/ushahidi/platform/issues/4908). It prevents weak passwords (despite achieving minimum length requirements) from being able to create accounts. It does this by disabling the signup button.

## How to Test this PR
1. Open the Ushahidi platform (either via a deployment or Localhost)
2. Click on `Login & Signup`.
3. Switch to the Signup view and fill in the form (except the password form).
4. For the password, fill in something that only meets the minimum length constraints but ignores other constraints (for example "adminooo"). Note: you can copy and paste the example.
5. You will notice that the password hint suggests that your password is very weak.
6. Agree to the terms and conditions.
7. You will notice that the **Signup** button is no longer disabled and can be clicked on.
8. Checkout to this PR.
9. Repeat steps 1-6.
10. You will notice that the **Signup** button is still disabled and can't be clicked on.
11. Try updating the password to meet another constraint such as adding a number. For example "adminooo1".
12. You will notice that the **Signup** button is no longer disabled and can be clicked on.

### DIFF
--- a/apps/mobile-mzima-client/src/app/shared/components/password-strength/password-strength.component.ts
+++ b/apps/mobile-mzima-client/src/app/shared/components/password-strength/password-strength.component.ts
@@ -88,7 +88,7 @@ export class PasswordStrengthComponent implements OnChanges {
     force += 2 * password.length + (password.length >= 10 ? 1 : 0);
     force += passedMatches * 10;
 
-    force = password.length <= 6 ? Math.min(force, 10) : force;
+    force = password.length < 8 ? Math.min(force, 10) : force;
 
     force = passedMatches === 1 ? Math.min(force, 10) : force;
     force = passedMatches === 2 ? Math.min(force, 20) : force;
--- a/apps/web-mzima-client/src/app/auth/forms/registration-form/registration-form.component.html
+++ b/apps/web-mzima-client/src/app/auth/forms/registration-form/registration-form.component.html
@@ -42,7 +42,10 @@
       }}</mat-error>
       <mat-hint>{{ 'user.password_hint' | translate }}.</mat-hint>
     </mat-form-field>
-    <app-password-strength [passwordToCheck]="form.value.password"></app-password-strength>
+    <app-password-strength
+      [passwordToCheck]="form.value.password"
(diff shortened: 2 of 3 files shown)


Chiemezuo:
@Angamanga I'd like your thoughts on this.

ushahidlee (changes requested):
@Chiemezuo Thank you for contribution!

@Angamanga Initially this works, but once you have typed in a strong password you can then delete the characters back to anything above a length of 6 and the button is still enabled.

Chiemezuo:
@ushahidlee I just checked this, and it isn't the case with this PR.
I started with a password of `admin` (5 letters) and the button was disabled.
Then I changed to a password of `adminooo` (8 letters) and the button was disabled, because it still had only alphabets.
Afterwards, I changed the password to `admin123` (8 letters with numbers) and the button was active.
Again, I deleted and changed it to `admin000` (8 letters) again, and the button was disabled again.


I will try to show you a screen recording in the next comment.

Chiemezuo:
Here's the video @ushahidlee 
https://github.com/ushahidi/platform-client-mzima/assets/29470516/28742ffb-c46a-4324-b758-b44aa1f7de57

PR #992, opened 2024-03-31
USH-1253: fixate humanitarian layer URL scheme

tuxpiper:
Upon debugging in safari connected to iOS simulator, I noticed that the tile requests are prefixed by "capacitor://"

This change makes the base layer scheme absolute.


ushahidlee (approved):
Looks good

Issue #724, opened 2023-12-12
Survey Configuration - Hide exact location information


theobadha:
Posts with location information have the location of the post shown on the map. By default the option to hide the location information is disabled. In this test, we will enable hiding the post location and verify that the exact location information is not shown on the map.

Test Steps

- [x]  As an admin, while logged in, open a survey in edit mode Settings -> Surveys -> Click on a Survey
- [x]  Click on Configure Tab
- [x]  Hide author information is disabled by default
- [x]  Click toggle switch to enable Hide Location information
- [x]  Save
- [x]  Log out as admin
- [x]  Log in as a different registered user
- [x]  Submit a post to the survey above
- [x]  Verify that the seen post location is not an exact match to the location provided when creating post

**We may need to use screenshots here to verify location data on the map.

PR #997, opened 2024-04-02
E2e tests for survey configuration hide exact location information

shakirandagire:
@AmTryingMyBest Could you review this PR for[ issue](https://github.com/ushahidi/platform-client-mzima/issues/724)


theobadha (approved):
I made several changes in this PR, but noteworthy is: 
- instead of the screenshot plugin and comparing hidden location using a screenshot, i opted to use checking lat and long. values instead as done for the other configuration tests

PR #999, opened 2024-04-03
Removing duplication of subcategories when creating and editing a post

Angamanga:
This PR removes a bug that caused subcategories to show up twice when creating and editing a post. 
To test:
Start add a post to a survey with a category-field with categories and subcategories.
- [ ] The subcategories should only appear once, under the parent category

Repeat above when editing a post.


ushahidlee (approved):
Looks good and does what its supposed to

PR #1007, opened 2024-04-09
Ush 1193 - Filters should behave

ushahidlee:
**Problem**:

The filters being sent in requests to our backend API are inconsistent and in some cases nonsensical. Also, they often contain too much information, resulting in slower queries. 

- map view might request unstructured posts (which dont have any geo data),  (field: _include_unstructured_posts_)
- data view might request geolocated posts. (fields: _has_location_, _within_km_, _include_unmapped_)
- forms might include 0 (which represents "unknown form" on the frontend but has no meaning in the backend)
- arrays in the request might be included in the request even if empty.
- selected forms may alter when changing views

**Tickets**:

https://linear.app/ushahidi/issue/USH-1193/inconsistent-behavior-of-unknown-form-checkbox-and-include
https://linear.app/ushahidi/issue/USH-1194/unintended-visibility-of-unknown-form-checkbox-in-map-view
https://linear.app/ushahidi/issue/USH-1195/inconsistent-include-unstructured-posts-parameter-value-upon

**Solution**:

This PR attempts to move all the filter logic into the service, meaning any logical changes to filters are happening just before the request is sent. With all the logic in one place, it also attempts to remove meaningless fields from the filters.

**Testing**:

As this PR alters everything to do with filters, all filter behaviour should be observed while testing. However, the requests sent to the backend should be the specific focus of testing, with the following conditions:

All Requests:
- No empty arrays.
- _include_unstructured_posts_ should be present in every request.

Map View
- _include_unstructured_posts_ should always be false.
- _has_location_ should always be 'mapped'.

Data View
- _include_unstructured_posts_ should be true if "unknown form" is selected, false otherwise.
- _has_location_, _place_ and _within_km_ shouldnt be in the request.

In addition, each view has 2 requests associated with it. The above conditions should apply to both requests.

Map View
/api/v5/posts/stats
/api/v5/posts/geojson

Data View
/api/v5/posts/stats
/api/v5/posts

### DIFF
--- a/apps/web-mzima-client/src/app/feed/feed.component.ts
+++ b/apps/web-mzima-client/src/app/feed/feed.component.ts
@@ -41,7 +41,7 @@ export class FeedComponent extends MainViewComponent implements OnInit {
   public override params: GeoJsonFilter = {
     limit: 20,
     page: 1,
-    include_unstructured_posts: false,
+    currentView: 'feed',
     // created_before_by_id: '',
   };
   private readonly getPostsSubject = new Subject<{
--- a/apps/web-mzima-client/src/app/feed/feed.component.ts
+++ b/apps/web-mzima-client/src/app/feed/feed.component.ts
@@ -41,7 +41,7 @@ export class FeedComponent extends MainViewComponent implements OnInit {
   public override params: GeoJsonFilter = {
     limit: 20,
     page: 1,
-    // include_unstructured_posts: true,
+    include_unstructured_posts: false,
     // created_before_by_id: '',
   };
   private readonly getPostsSubject = new Subject<{
--- a/apps/web-mzima-client/src/app/feed/feed.component.ts
+++ b/apps/web-mzima-client/src/app/feed/feed.component.ts
@@ -41,7 +41,7 @@ export class FeedComponent extends MainViewComponent implements OnInit {
   public override params: GeoJsonFilter = {
     limit: 20,
     page: 1,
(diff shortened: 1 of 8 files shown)


Angamanga:
A few findings when testing this:
- The selectAll-button seems broken
- Sometimes, the result-numbers are 0, and sometimes they are not, for surveys:
- Screenshot 2024-04-10 at 09 38 55
 It seems to happen after clicking "clear" and then "select all". Clearing the filters make it go back to normal again.
- When there is no data-sources connected to a deployment, there is still "Unknown form" in the filters-list. This is confusing to users and Cecilia has been asking about removing it but it may be out of scope for this ticket?
- All the rest seem to be working fine! 👍

ushahidlee:
> A few findings when testing this:
> 
> * The selectAll-button seems broken

Fixed.

> * Sometimes, the result-numbers are 0, and sometimes they are not, for surveys:
> * Screenshot 2024-04-10 at 09 38 55
>   It seems to happen after clicking "clear" and then "select all". Clearing the filters make it go back to normal again.

Fixed.

> * When there is no data-sources connected to a deployment, there is still "Unknown form" in the filters-list. This is confusing to users and Cecilia has been asking about removing it but it may be out of scope for this ticket?

So this is outside the scope of this ticket, but this PR does remove "Unknown Form" from the map view as it never makes sense there.

> * All the rest seem to be working fine! 👍

Angamanga (approved):
All fine now 🎉

PR #1008, opened 2024-04-11
Adds hidden field functionality to data source settings

ushahidlee:
**Issue**

There is no way to persist fields in data source settings that arent user-modifiable. For example. the email data source needs to store the id of the last retrieved email to prevent it being imported multiple times. 

**Solution**

This PR adds a field type of 'hidden' to the data source config form, allowing the system to hide config values from the user.

**Testing**

- Data Source config forms appear and behave correctly.
- The email data source form has an input of type=hidden in it called "incoming_last_uid".


tuxpiper (approved):
nice 👍

tuxpiper:
Heads up though that e2e-tests are red

PR #1009, opened 2024-04-12
Ush 1220 - After initial load, the mobile app does not switch to a different tile layer on the map

ushahidlee:
**Issue**:

After initially loading, the mobile app does not switch tile layers on the map even if a deployment is selected that would necessitate it.

**Solution**:

This PR alters the leaflet map handling code to allow for tile layers to change after the initial load.

**Testing**:

1. Open the mobile app
2. Note the tile layer
3. Select a different deployment with a different tile layer.
4. Note the changed tile layer

### DIFF
--- a/apps/mobile-mzima-client/src/app/map/components/map-view/map-view.component.ts
+++ b/apps/mobile-mzima-client/src/app/map/components/map-view/map-view.component.ts
@@ -73,39 +73,35 @@ export class MapViewComponent implements AfterViewInit {
     this.sessionService.mapConfig$.subscribe({
       next: (mapConfig) => {
         if (mapConfig) {
-          console.log('LOGGY - In the Map Config Listener');
-          this.mapConfig = mapConfig;
-          console.log('LOGGY - Map Config: ' + JSON.stringify(this.mapConfig));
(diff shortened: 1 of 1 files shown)


apps/mobile-mzima-client/src/app/map/components/map-view/map-view.component.ts
```diff
@@ -73,22 +73,37 @@ export class MapViewComponent implements AfterViewInit {
...
+              });
+              this.mapConfig = mapConfig;
+              this.baseLayer = this.mapConfig.default_view!.baselayer;
+              const currentLayer = mapHelper.getMapLayer(this.baseLayer, this.isDarkMode);
+              this.offlineLayer = tileLayerOffline(currentLayer.url, currentLayer.layerOptions);
+              this.onlineLayer = tileLayer(currentLayer.url, currentLayer.layerOptions);
+              this.map.addLayer(this.offlineLayer);
+              this.map.addLayer(this.onlineLayer);
+            }
+          }
```
Ifycode, line 106: Works fine as in the issue ticket
> Although there is a failing test, which I can't tell it it's random or related

Ifycode (approved):

PR #1014, opened 2024-04-15
USH-1266 - Deployment map + data still visible after Deletion

ushahidlee:
**Issue**: 

- The deployment map and data are still visible and active after deleting the deployment.
- The deployment name is no longer visible in the deployment list but is still visible within the profile
- The map data only changes after the user switches deployment.
- If the user switches the deployment, the deployment name and data switch correctly
- If the app is uninstalled and reinstalled, all deleted deployments reappear.

**Expectation**:

- Deployment name should not be visible in the deployment list or profile after successful deletion
- Deployment map + data should not be visible after deletion
- After deleting a deployment, the user should be signed out of deployment if previously logged in

**Testing**:

1. Go to the deployment list and Delete 1 or more deployments
2. Close or uninstall the app
3. Reopen or reinstall the app
4. Access the map view to see if the deleted deployment is still visible 
5. Access the deployment list to see if deleted deployments are still visible
6. If previously logged in to the deleted deployment, check if the user is still logged in after the deletion


Angamanga (approved):
All fine 🎉 Extra bonus for removing old commented out stuff :D

PR #1020, opened 2024-04-17
update login credentials

theobadha:
This change from having the e2e tests as secrets to having them out in the open was necessitated by the fact that contributors would have the tests failing in (external) forks. This change will allow tests from any forks to run with the correct credentials.


theobadha:
https://linear.app/ushahidi/issue/USH-894/remove-secret-protection-for-e2e-tests-variables

PR #1021, opened 2024-04-17
Checking for correct permission-values for collections

Angamanga:
Adds an extra check for the correct values for the role 'everyone'.

Fixes USH-1247

To test USH-1247:
1. Click on the "Collections" section.
2. Click the "Add Collection" button.
3. Create a new collection with the visibility of "Everyone" 
4. Click on the "Collections" section.
5. Search for the newly created collection.
6. Click on the edit icon of the newly created collection.
- [ ] The visibility should still be "Everyone"
8. Change to any other role/s and save
9. Go back to edit the collection
- [ ] The selected roles should still be selected

To test USH-1272
10. Select "Everyone" for above collection again
11. Logout
12. Click on the "Collections"-section
13. The collection should be visible in the list.


ushahidlee (approved):
Works and looks great

PR #1029, opened 2024-04-19
remove stand-alone tests that verify posts review before publishing

theobadha:
@shakirandagire in this commit, I removed the stand-alone test that checked for the permission to review a post before it is published. I did this because this step or action is also carried out in the test that already exists for checking the hide author.
I figure since that action is already checked in another test, no need to have more code that repeats a test. Less code is good code. We may however bring it back should we realize the need to. 
LMK what you think.

### DIFF
--- a/e2e-testing/cypress/e2e/13-survey-configurations/survey-configurations.cy.js
+++ b/e2e-testing/cypress/e2e/13-survey-configurations/survey-configurations.cy.js
@@ -10,10 +10,6 @@ describe("Automated Tests for Survey Configurations", () => {
     cy.visit(Cypress.env('baseUrl'));
   });
 
-  it.skip("Verify posts go into Published state as configured", () => {
-    surveyConfigurationFunctions.require_posts_reviewed_before_published()
-  });
-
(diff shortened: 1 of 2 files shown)


shakirandagire (commented):
@AmTryingMyBest It looks good to me. I agree with the less code aspect. I just have an inquiry about where we should check if the post has published status or not

theobadha:
> @AmTryingMyBest It looks good to me. I agree with the less code aspect. I just have an inquiry about where we should check if the post has published status or not

The test to check for author hidden has a step where logout happens. When logged out, the post is visible to logged out users since it's published, so the publish status is taken care of.

shakirandagire:
> > @AmTryingMyBest It looks good to me. I agree with the less code aspect. I just have an inquiry about where we should check if the post has published status or not
> 
> The test to check for author hidden has a step where logout happens. When logged out, the post is visible to logged out users since it's published, so the publish status is taken care of.

That makes sense. Thanks for the response.

shakirandagire (approved):
Looks good to me

PR #1030, opened 2024-04-19
Filter Undefined Issue

ushahidlee:
**Issue**:

Under certain conditions the filters will create an "undefined" limit value causing the request to fail.

**Solution**:

Add a sanity check at the end of the filter processing, but before sending the request, that deletes all undefined values from the filter.

**Testing**:

1. Open map view
2. Clear filters
3. Clear all selected forms
4. See results on map


Angamanga (approved):
👣 👣

PR #1036, opened 2024-04-25
USH-1280 Remove/Amend incorrect filter parameters for posts 

ushahidlee:
**Issue**:
Some API calls coming from the mobile app had incorrect parameters that were resulting in errors.

**Ticket**:
https://linear.app/ushahidi/issue/USH-1280/remove-unnecessary-parameters-when-call-postsgeojson-from-mobile

**Solution**:
The mobile app now communicates its state better to the shared service and erroneous parameters have been removed or corrected.

**Testing**:
The requests coming from the app should not have the following parameters:

1- order
2- orderby
3- order_unlocked 
4- reactToFilter

also, limit was 100000 but has been modified to 500 as it is in the web client.


Angamanga (approved):
Looks good and works fine. I tried a few different filters and I could see none of the parameters mentioned in the description 👍

PR #1037, opened 2024-04-25
Select all children when selecting a parent in Settings USH-887

Angamanga:
This PR automatically selects/deselects and disables children when a parent is selected when activating Bulk-actions in Settings.

To test
- Login and go to settings
- Select categories and activate "Bullk actions"
- [x] All children should automatically be expanded 
- Select a parent
- [x] All its children should automatically be selected
- [x] All the checkboxes for the children should become disabled
- Deselect a parent
- [x] All its children should automatically be de-selected
- [x] All the checkboxes for the children should become enabled again

- [x] Children can be deleted without deleting the parent
- [x] Parents cannot be deleted without all its children
- [x] Categories without children can be deleted

Bonus-fix: Adjustment of label-placement when bulk-actions is activated


ushahidlee (approved):
Looks and works great 👍

PR #1040, opened 2024-04-27
Map pin duplication bug & 

ushahidlee:
**Issues**:

This PR deals with 2 issues:

1. The Map View makes multiple calls to the geojson endpoint, and uses the response twice resulting in duplicated pins(posts) on the map.
2. Under specific conditions if a user selects no surveys, they will instead get all the posts from every survey returned.

Unfortunately the fix for the second issue built upon the changes in the first, which is why the issues are in one PR.

**Tickets**:

https://linear.app/ushahidi/issue/USH-1315/duplication-of-pins-in-map-view
https://linear.app/ushahidi/issue/USH-1275/incorrect-filtering-all-survey-forms-returned-when-no-survey-option

**Solutions**:

1. This is due to the way we are handling state management in our frontend. While that definitely needs to be addressed, this PR attempts to understand the state of the client to understand if it should use the response of the API call or ignore it. There is some voodoo involved in handling pagination, so I apologise in advance to the reviewer.
2. This is a relatively simple fix, sending 'none' in the form array.

**Testing**:

1.
    a) Open the Map View
    b) Check clusters of pins to see if there are any duplicates.

2. 
    a) Open the Map View.
    b) Clear the form filter. The map should have no pins on it.
    c) Select a specific form and then deselect it. The map should again have no pins on it.

### DIFF
--- a/apps/web-mzima-client/src/app/core/helpers/search-form.ts
+++ b/apps/web-mzima-client/src/app/core/helpers/search-form.ts
@@ -18,6 +18,8 @@ export const statuses = [
   },
 ];
 
+export const loggedOutStatuses = [statuses[0]];
+
 export const sources = [
   {
     name: 'Web',
@@ -116,7 +118,7 @@ export const sortingOptions = [
 
 export const DEFAULT_FILTERS = {
   query: [''],
-  status: [['published', 'draft']],
+  status: [['published']],
   tags: [[]],
   source: [[]],
   form: [[]],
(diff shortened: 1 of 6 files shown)


Angamanga (approved):
This works in branch ush-1181. Clearing of the filters however is returning a bit random results, but I *think* that is another issue.

Angamanga:
Closing in favour of https://github.com/ushahidi/platform-client-mzima/pull/1041

PR #1052, opened 2024-05-08
Displaying checkmarks for correct collections in map-view

Angamanga:
This updates the post-object in map-view so selected collections are visible when opening the collection-modal again.
To test:
As a logged in user:
- Go to map-view
- Click on a post
- Click on "..."
- Click on "Add to collection"
- Select a number of collections by checking them in the modal (and remember which ones)
- Close the modal
- Reopen the modal without refreshing the page
- [x] The same collections are checked

Bonus thing: A done-button in the bottom of the modal.


ushahidlee (approved):
Works well :+1:

PR #1056, opened 2024-05-09
[USH-1180] Remove Post actions option for unlogged in users

MMwandigha:
I removed the Post Actions button for users who are not logged in as per the ticket,
Only logged-in users can see the post actions button (three-dot button)


apps/mobile-mzima-client/src/app/map/components/post-item/post-item.component.html
```diff
@@ -20,7 +20,7 @@
       fill="clear"
       shape="normal"
       color="medium"
-      *ngIf="!checkbox"
+      *ngIf="isLoggedIn && !checkbox"
       class="post__details"
```
Angamanga, line 24: @MMwandigha You could solve the flicker of the dots with creating a function checking the status directly instead of assigning a variable in `onInit`. Something like this:
```
public getLoggedInStatus() {
    return this.sessionService.isLogged();
  }
```

In that way, it would check immediately instead of waiting for the variable to be assigned/reassigned, I think thats what causes the flicker.

Angamanga (changes requested):

apps/mobile-mzima-client/src/app/map/components/post-item/post-item.component.html
MMwandigha: @Angamanga I see, let me try doing this instead

Angamanga (approved):
All good! Thank you for this 🎉

PR #1059, opened 2024-05-11
Ush 1340 - No map and no posts when site loaded with no local storage

ushahidlee:
**Issues**:

On loading the site with no cache or local storage (eg. an Incognito Window), the map does not load and also no pins.

**Solution**:

1. The filter logic had assumed there would always be filters stored in local storage and would erase the defaults with null. Obviously this isnt always the case, so I have added a check to prevent this happening.
2. The logic that parses the filters immediately before sending to the API was assuming no forms selected for the above circumstance, even though there were forms selected.

**Testing**:

1. Open an incognito window
2. Open the site.
3. The map loads and also loads the pins for the selected deployment.
4. Refresh the window
5. The map again loads and also loads the pins for the selected deployment.


Angamanga (approved):
Makes sense! Approved!

PR #1064, opened 2024-05-15
[USH-1328] All surveys are listed in mobile, no matter what permissions the user have

MMwandigha:
**Actual behaviour**
The survey is visible, the user can tap on the survey and fill in the form, but when trying to tap "submit" you get an "unauthorised" error message

**Expected behaviour:**
The survey with the specific permissions should not be visible in the survey-list

### DIFF
--- a/apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.ts
+++ b/apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.ts
@@ -328,8 +328,23 @@ export class PostEditPage {
     this.updateFormControl(key, dateHelper.setDate(event.detail.value, type));
   }
 
-  private async loadSurveyFormLocalDB() {
-    return this.dataBaseService.get(STORAGE_KEYS.SURVEYS);
+  private async loadSurveyFormLocalDB(): Promise<any[]> {
+    try {
+      const surveysFromDB: any[] = await this.dataBaseService.get(STORAGE_KEYS.SURVEYS);
+      const filteredSurveys = surveysFromDB.filter((survey) => {
+        return (
+          survey.everyone_can_create ||
+          survey.can_create.includes(
+            localStorage.getItem(`${generalHelpers.CONST.LOCAL_STORAGE_PREFIX}role`),
+          )
+        );
+      });
+
+      return filteredSurveys;
+    } catch (error) {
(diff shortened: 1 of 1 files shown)


apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.ts
```diff
@@ -316,8 +328,23 @@ export class PostEditPage {
...
+          )
+        );
+      });
+
+      return filteredSurveys;
+    } catch (error) {
+      console.error('Error loading surveys from local database: ', error);
+      return this.dataBaseService.get(STORAGE_KEYS.SURVEYS);
+    }
   }
```
Angamanga, line 347: Hm, I think you need to remove the return in the catch on row 346. Otherwise it will just throw another error?

Angamanga: Since its catching an error, and then trying to do the same thing again.

apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.ts
```diff
@@ -189,8 +191,18 @@ export class PostEditPage {
...
+            survey.can_create.includes(
+              localStorage.getItem(`${generalHelpers.CONST.LOCAL_STORAGE_PREFIX}role`),
+            )
+          );
+        });
+
         await this.dataBaseService.set(STORAGE_KEYS.SURVEYS, response.results);
-        return response.results;
+        return filteredSurveys;
       } catch (err) {
```
Angamanga, line 206: All fine here 👍

Angamanga (changes requested):
See my comment in the code, just remove the return on 346 👍

apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.ts
MMwandigha: @Angamanga Aah I see. Let me fix that

MMwandigha: Can we return an empty array instead? Something like
```
console.error('Error loading surveys from local database: ', error);
    return [];
```

Angamanga (approved):
All good! Merging 🎉

PR #1071, opened 2024-05-20
Changing survey-list from dropdown to list

Angamanga:
Displaying a list of surveys instead of a dropdown when adding a post + some refactoring to make the list-style app-wide (we use them on many places)

To test:
In mobile-app, map-view
1. Tap on + sign to add a post
- [x] There is a list with surveys
2. Tap on a survey
- [x] The form for the survey is displayed


ushahidlee (approved):
Super nice experience!

PR #1072, opened 2024-05-20
Removing null from image-field instructions

Angamanga:
Removes the text "null" from the Image-field if there are no field-instructions available.
To test:
1. Create an image-field in a survey. Do not touch the default field-description
2. Try adding a post, observe the default-field description
3. Go back to survey-editor, remove the default field description
4. Try adding a post to the survey
 - [x]  there is no "null" in front of the text about Max image size
5. Go back to survey-editor, add a field description
6. Try adding a post to the survey
- [x] The formatting of the field description and Max image size-text is fine and not cut off anywhere


ushahidlee (approved):
Looks good and works well!

PR #1073, opened 2024-05-21
[USH-1355] The 'Set new Photo' button sometimes requires multiple clicks for file input to function correctly

MMwandigha:
**Ticket Description**
The profile photo sometimes requires the user to select the 'set new photo' button multiple times for the file input function to work correctly. Possible solution to this is resetting the file input after selection to ensure the input is ready for a new file

**Solution;**
Instead of relying on the previous label element to trigger the file input dialog, I created a separate function that programmatically triggers this. This function will then ensure the file input value is reset before the dialog is opened.


ushahidlee (approved):
Clever solution and works well

MMwandigha:
> Clever solution and works well

Great! Thanks @ushahidlee

PR #1074, opened 2024-05-22
USH-1347 - Deleted Posts do not get removed in Map View

ushahidlee:
**Issue:**

When a post is deleted on the map view, the pin on the map persists. For the map to be updated, the page has to be reloaded.

**Solution:**

When a post is deleted, it should be removed from the map without the need to reload the page for the map(or the posts on the map) to be updated.

**Testing**:

1. Log into site as admin
2. Goto map view
3. Click on a post and delete it
4. Post will be removed from the map and the totals updated.


Angamanga (approved):
Pin is gone after deleting! Nice! Approved!

PR #1086, opened 2024-05-31
[USH-942] Minor adjustment to margin

MMwandigha:
Just added a very slight margin below the language selector for uniformity

Before:
image

After:
image


Angamanga (approved):
Looks good now 🙌 I saw that the dropdown is a bit off though, there string is not translated properly. I will create a ticket for it and you can work on that one as well (its a small fix 🎉 )

MMwandigha:
> Looks good now 🙌 I saw that the dropdown is a bit off though, there string is not translated properly. I will create a ticket for it and you can work on that one as well (its a small fix 🎉 )

@Angamanga No problem. Thanks!

PR #1088, opened 2024-05-31
USH-1365 - Posts do not appear in 'My Posts' page on the mobile app

ushahidlee:
**Issue**:

In the mobile app, My Posts always returns 'Posts not found'.

**Solution**:

App was incorrectly filtering out all forms in the my posts query. It now passes 'myposts' as current view to the posts service which then creates the appropriate posts query.

**Testing**:

1. Open App
2. Select any deployment you created posts for
3. Login your credentials
4. Navigate to profile
5. Click on 'My Posts'
6. See correct listing of posts for the logged in user.


Angamanga (approved):
All good 👍

PR #1089, opened 2024-05-31
USH-1117 Screen doesn't scroll to accommodate keypad

ushahidlee:
**Issue:**

On the Android app, when the keyboard is opened the screen does not always scroll to accommodate the user being able to see the field. This is especially noticeable on the login page where you cannot see the password field at all while typing in it.

**Solution:**

There is an undocumented feature of the capacitor keyboard plugin specifically for Android to address this. I have added the plugin and the necessary configuration.

**Testing:**

1. Open the android app.
2. Select any deployment
3. Go to profile and then login
4. Click into the password field.
5. Notice that the screen scrolls to allow you to see the password field fully.

### DIFF
--- a/apps/mobile-mzima-client/capacitor.config.ts
+++ b/apps/mobile-mzima-client/capacitor.config.ts
@@ -34,6 +34,9 @@ const config: CapacitorConfig = {
         biometricSubTitle: 'Log in using your biometric',
       },
     },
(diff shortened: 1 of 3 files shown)


Angamanga (approved):
Works fine now, tested on Android 11

PR #1094, opened 2024-06-05
[USH-1183] App redirects to Profile page on deleting post from My Posts page

MMwandigha:
Fix for app redirection after bulk delete in 'My Posts'

**The Problem:**
Once you navigate to profile>my posts, select the posts you want to delete using 'bulk actions', and then delete them, the app redirects you to the profile page instead of staying on the same 'My Posts' screen.

**Expected behaviour.**
It should stay on that 'my posts' page and update automatically once bulk posts are deleted.

**Approach:**

- Emitting the event with the list of the deleted post IDs for the 'handlePostDeleted' method in `posts.page.ts` so as to remove them from the displayed list of posts
- I also added the function `this.isEditMode = false; ` to revert the bulk selection mode after deleting

### DIFF
--- a/apps/mobile-mzima-client/src/app/profile/posts/posts.page.html
+++ b/apps/mobile-mzima-client/src/app/profile/posts/posts.page.html
@@ -40,7 +40,7 @@
     [permissions]="permissions"
     [isProfile]="true"
     (postChanged)="getMyPosts()"
-    (postDeleted)="back()"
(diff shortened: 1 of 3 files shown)


MMwandigha:
@ushahidlee Could you have a look at this?

ushahidlee (approved):
Works well and is a tidy solution :+1:

PR #1095, opened 2024-06-05
USH-1397 Filters do not work correctly in mobile app

ushahidlee:
**Issue**:

The posts being returned to the user do not reflect the values in the filters.

**Solution**:

This PR slightly modifies the logic in the posts service to manipulate the filter values before sending the request to the API.

**Testing**:

1. Click on filters to reveal filters page
2. Note the post count (top right)
3. Make a change to a filter option for instance, click on Status and select a different combination of status
4. Click Apply to effect changes
5. Note that the post count changes to reflect the new filtered dataset.
6. In data view, the posts returned match the filtered dataset.


Angamanga (approved):
All good now 🌮

PR #1096, opened 2024-06-06
[USH-738] Surveys: select_language value appears for specific surveys as default translation

MMwandigha:
Description: "select_langauge" and "(default)" values appear for specific surveys as default translation

When there are no translations selected, the "English(default)" value should be automatically present instead in both dropdown title and list. (issue is present for older surveys)


apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -94,7 +100,6 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
 
   private initLanguages(enabledLanguages: SurveyItemEnabledLanguages) {
     this.languages = this.languageService.getLanguages();
-    this.defaultLanguage = this.languages.find((lang) => lang.code === enabledLanguages.default);
```
Angamanga, line 97: This is still needed in case someone has set another language as default, not english.

Angamanga (approved):
Excellent! This now works fine! Could you do the same thing for when adding a post? That is fixed in post-edit.component.ts.
It could be done in this same branch 👍

MMwandigha:
> Excellent! This now works fine! Could you do the same thing for when adding a post? That is fixed in post-edit.component.ts.
> 
> It could be done in this same branch 👍 

Thanks Anna, for sure let me do that👍

Angamanga (approved):
All good now :)

PR #1097, opened 2024-06-09
Adding coordinates as text instead of input-boxes

Angamanga:
This PR removes the textboxes where the lat and long for the map pin was displayed and shows them as uneditable text instead.

To test:
In mobile app,
- Add a post with a location-field
- Search for a location and select one from the results
- [ ] The coordinates is visible as text above the map
- Move the pin
- [ ] The coordinates changes to the new location

Design: https://www.figma.com/design/c0LFvzuib8E3FHmp6Of8MS/Ushahidi?node-id=15182-130747&t=8yjwl2fX6vHjFKtc-0 
Note: The colour used for the coordinates is slightly different, I used a colour we already used and has defined in the app.


ushahidlee (approved):
looks great!

PR #1099, opened 2024-06-09
[USH-1401] Status filters on Android show Draft and Archived statuses for logged out user.

MMwandigha:
**Describe the bug**
Status filters on Android show Draft and Archived statuses for logged out user.

**Where was the bug observed**
Android mobile application

**To Reproduce**

- Go to mobile application
- Log out of application
- Click on Filters icon.
-  Click on Status, notice that the Draft and Archived are shown instead of just Published like on the web.

**Expected behavior**
Status filters on Android do not show Draft and Archived statuses for logged out user.


MMwandigha:
Current Challenges faced here:
1.  `'selectedCount'` only changes if you refresh the page, but at least shows 1 for loggedOut and 3 for loggedIn. I'm not sure why if it is a problem with initialization of the component.
2.  `'selected'` is where the trouble comes:
- If I was logged in and selected both the three options, the `'selected'` count still shows three after logging out and refreshing the page to confirm if there are changes.
- If I select 'clear' to clear all filters, the count reverts to one which is how it should behave initially

If only the selected options would dynamically change based on whether the user is loggedIn or not, that would be ideal.

PR #1110, opened 2024-06-15
[USH-1407] Disabled "Set Image" Button After Media Update Error pen_spark

MMwandigha:
(Still in Draft)
Problem: 

When an error occurs while updating the media associated with a post or profile, the "Set Image" button remains disabled and all page will be stoPped until uploading image is finished. This behavior can be confusing for users who might not understand why they cannot proceed after encountering an error.

Expected behavior:

The "Set Image" button should provide clear feedback based on the media update status

- [x] Successful Update: If the media update is successful
- [x] Update Error: Display a clear error message explaining the issue, consider enabling the button once the user has addressed the error (e.g., by selecting a different image file) 
- [x] * cancel and back button must keep working

Steps to Reproduce:

1- login with user

2- click on profile 

3- edit user profile

4-update the profile image with non-image file

5- check the results


ushahidlee (approved):
A nice user experience, and the code looks good! :+1:

PR #1124, opened 2024-06-20
Displaying privacy policy and t&c properly

Angamanga:
This PR uses innerHTML to display the Privacy Policy and T&C that contain html-tags.
To test:
- Go to profile-page
- Click on "Terms and Conditions" and "Privacy policy"
- [x] They should be displayed without html-tags.


ushahidlee (approved):
Tidy :+1:

PR #1125, opened 2024-06-20
Hiding language-switcher for now

Angamanga:
This PR hides (comments out ) the language-switcher until we have more translations available.

To test: 
In mobile,
- Go to the Profile menu
- [ ] No language-switcher should be visible


ushahidlee (approved):
:+1:

PR #1128, opened 2024-06-24
Add other-option when configuring a survey

Angamanga:
Adding configuration-option for "Other" in the survey-fields radio, checkboxes and select.
To test:

In survey-settings, add a new field for Radio and Checkboxes, in all those fields, test the following:
1. Before any options are added, the button "Add other" is not visisble
2. Click on "Add +"
- [x]  An empty input-field appears where the user can add an option and there is also a button "Add other"
3. Click on "Add other"
- [x] A disabled input-field appears with the value "Other"
- [x] The "Add other" button is not visible
4. Click on the trash-can for the other-option
- [x] The option is removed
- [x] The "Add other" button is visible again

-

### DIFF
--- a/apps/web-mzima-client/src/app/settings/surveys/create-field-modal/create-field-modal.component.html
+++ b/apps/web-mzima-client/src/app/settings/surveys/create-field-modal/create-field-modal.component.html
@@ -87,7 +87,9 @@
         </div>
         <div
           *ngIf="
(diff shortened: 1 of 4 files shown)


ushahidlee (approved):
Nice and the code is good too. The PR mentions this should be available for "select" fields. Its not available, but the original ticket doesnt mention that so I'm approving it :+1:

PR #1134, opened 2024-06-28
[USH-1246] Inconsistent display issue on the home/explore screen for mzima mobile app

MMwandigha:
Describe the bug
On the home/explore screen of the mobile app, there appears to be some experience inconsistency, that might prove to be uncomfortable to the users.

Where was the bug observed?
Mzima Mobile (Home Screen)

Expected behavior

The menu/toolbar shouldn't disappear whenever the feed view overlay is scrolled to the end.

The "add post" and "reload posts" buttons should be static at the position.


MMwandigha:
Found a slight error. Moving it back to 'In Progress"

MMwandigha:
All good now

apps/mobile-mzima-client/src/app/map/map.page.html
```diff
@@ -2,7 +2,7 @@
   #layout
   [mode]="1"
   [offsetTop]="78"
-  [breakpoints]="[0, 300]"
+  [breakpoints]="[78, 300]"
```
Ifycode, line 5: "Senior Intern Engineer" 🙌🏽 Nice figuring this part out @MMwandigha 

Although when you scroll back down it bounces (goes hidden behind the menu) like once or twice before it now stays visible above the menu.

Maybe you want to make it more smooth, so that it just takes its place straight once scrolled down.

Ifycode (changes requested):

apps/mobile-mzima-client/src/app/map/map.page.html
MMwandigha: 😅Thanks  @Ifycode for checking out the PR.
For sure, let me have a look at it and see how to optimize it as stated

Ifycode: Dropping extra note that something like that happens for the sorting area above the post, when you scroll and it grows to the top @MMwandigha (Although your "eagle" eyes may have already detected it 😉).

MMwandigha: @Ifycode Hmm I see. Do you suggest we remove the bouncing effect for both scenarios?

Ifycode: Yes @MMwandigha

MMwandigha:
@Angamanga You can review this

Angamanga (approved):
Neat! Works well now 🎉

PR #1127, opened 2024-06-24
Fixed keyboard focusability on the company-info section

kivuvarosekivuvan:
## Description
This PR addresses the issue with keyboard-inaccessibility of the company-info section. 

Fixes: https://github.com/ushahidi/platform/issues/4961

## Changes
-  Fixed the keyboard-accessibility of this section by using the tab index key

## Screen Walkthrough
[Screencast from 24-06-2024 01:42:55 ALASIRI.webm](https://github.com/ushahidi/platform-client-mzima/assets/124133577/6dc7b509-b87e-4d30-ba58-e1165e8ab8c3)

## Testing
- Tested manually and using a screen reader to read out the purpose of the image and ensuring there is no redudancy

PR #1135, opened 2024-07-01
Improve accessibility of company info expansion panel by adding keyboard focus handling and visual indicators

### LINKED ISSUE #4961: [Accessibility] Keyboard Inaccessibility of "Company-info" Section (Non-Focusable)
## Describe the bug
- The "Company-info" section in the top left of the interface is not keyboard-focusable. This makes it difficult for keyboard-only users to determine if they are currently positioned within that specific section of the UI, reducing their awareness of their location within the Ushahidi Platform

## To Reproduce
1. Launch the local host
2. Attempt to focus on the "Company-info" section using only the keyboard.
3. Observe that the "Company-info" section cannot be focused on or highlighted using the keyboard.

## Expected behavior
The "Company-info" section should be fully keyboard-accessible, allowing users to navigate to and focus on that part of the UI using only the keyboard. This would improve the our platform's usability for keyboard-only and assistive technology users.

## Screenshots
[Screencast from 19-06-2024 12:02:41 ASUBUHI.webm](https://github.com/ushahidi/platform-client-mzima/assets/124133577/77c9d3e5-74bc-4eb9-a45a-af7cbe862705)

kivuvarosekivuvan:
## Description
This PR addresses the issue with keyboard-inaccessibility of the company-info section. 

Fixes: https://github.com/ushahidi/platform/issues/4961

This is a change from the merged PR that addressed the same: https://github.com/ushahidi/platform-client-mzima/pull/1127


kivuvarosekivuvan:
### @Ifycode Hey 👋,
I've found a couple of things while working on this:

**1. Focus Issue with Open Accordion:** With the current changes, the focus appears to be on the header once the accordion is open. This means the focus outline is still not visible. I've tried conditionally applying CSS styles to show or hide the focus outline based on the input method using ngClass but still the same issue.

**2. Alternative Approach with focus-within:** Alternatively, we can achieve the desired behavior by directly applying the focus styles on the .company-info element using the :focus-within pseudo-class. This approach seems to work, but it may require some minor adjustments to the styling to get the desired look and feel. I have also tried this approach

```
.company-info {
  &:focus-within {
    outline: 1px solid #ffffff;
  }
}
```

The second approach using **:focus-within** seems like a more straightforward solution, since it'll allow us to apply the focus styles directly on the element. However, I really think this will need us to fine-tune the styling to achieve what we have on the other focus rings.

Please let me know what you think of this issue

apps/web-mzima-client/src/app/shared/components/company-info/company-info.component.scss
```diff
@@ -113,6 +113,16 @@
...
 
+.no-outline:focus,
+.no-outline:hover {
+  outline: 1px solid rgb(255, 255, 255);
+}
+
+.has-outline:focus,
+.has-outline:hover {
+  outline: 1px solid rgb(255, 255, 255);
+}
```
Ifycode: @kivuvarosekivuvan based on the findings you wrote in our slack chat (and now that I think about it more), we don't need a new class-name in the css. We also don't need all those javascript code. You can happily remove all the code you have added.

I've also seen your comment here on using focus-within inside of the company-info class. I think it may be more helpful and clear to developers if we target the class from the library/framework directly.

I'll suggest a more straightforward fix using only css. At the bottom of this `company-info.component.scss` file, you will find the `.mat-expansion-panel` class.

**_Step 1:_** You can restore/fix the **_focus_** of the `.mat-expansion-pane-header` from within here.

Now, after doing this, most parts of the focus ring on the header is likely to be hidden. This is because `mat-expansion-panel` element itself has some hidden overflow applied to it.

**_Step 2:_** You can also just restore/fix the overflow of the `mat-expansion-panel` element from within there also.

Really, that's all.

Ifycode (changes requested):
Hmm...

apps/web-mzima-client/src/app/shared/components/company-info/company-info.component.scss
```diff
@@ -116,9 +116,13 @@
 .mat-expansion-panel {
   &:not(.mat-expanded):not([aria-disabled='true']) {
     .mat-expansion-panel-header {
+      &:focus {
+        outline: 1px solid #fdfdfd;
+      }
```
Ifycode: Because it's inside `&:not(.mat-expanded):not([aria-disabled='true'])`, it would only work after the dropdown is expanded or so.
You can move it outside of it @kivuvarosekivuvan

apps/web-mzima-client/src/app/shared/components/company-info/company-info.component.scss
```diff
@@ -116,9 +116,13 @@
...
       &:hover {
         background-color: transparent;
       }
     }
   }
+  overflow: visible;
```
Ifycode: Move this to the top of this code block for easy readability @kivuvarosekivuvan

Ifycode (changes requested):

apps/web-mzima-client/src/app/shared/components/company-info/company-info.component.scss
Ifycode: Also let's use the browser default focus outline instead for consistency.

kivuvarosekivuvan:
Hoyee 🤩, @Ifycode Changes upto date 
cc: @Angamanga

apps/web-mzima-client/src/app/shared/components/company-info/company-info.component.scss
```diff
@@ -114,6 +114,14 @@
 }
 
 .mat-expansion-panel {
+  overflow: visible;
+
+  .mat-expansion-panel-header {
+    &:focus {
+      outline: auto;
```
Ifycode: Nice. Try `revert !important` instead of `auto` @kivuvarosekivuvan

Ifycode (changes requested):

Ifycode (approved):

PR #1140, opened 2024-07-03
[USH-1408] ALT - Inconsistent Submit Button Behavior on Profile Edit

MMwandigha:
**Problem:** 

When a user attempts to update their profile by only changing the profile image, the submit button remains disabled. This behavior prevents users from saving the profile update unless they modify additional fields, even if they don't intend to change anything else.

**Expected behavior:**

The submit button for profile editing should adapt dynamically:


MMwandigha:
Pending/Bugs (As of 03/07/2024)

- [x] The local storage of the photo has not been implemented yet, so image will disappear if the user doesn't save it
- [ ] The save button, while it works, is still buggy. Assuming we update the form and save the picture at the same time, the emitter from profile-photo.component is being called unnecessarily and there is a concurrent image upload _(_view the console behaviour)__
- [ ] Editing a form only also triggers the same behaviour

apps/mobile-mzima-client/src/app/profile/information/information.page.ts
```diff
@@ -136,8 +163,14 @@ export class InformationPage {
       });
   }
 
-  public handlePhotoChange(event: boolean): void {
-    this.isPhotoChanged = event;
+  public handlePhotoSelected(event: { key: string; caption: string }): void {
+    if (this.selectedFileKey !== event.key || this.selectedCaption !== event.caption) {
+      this.selectedFileKey = event.key;
```
Angamanga: I think `this.selectedFileKey` and `event.key` will always be the same, so this will never detect the new photo.

MMwandigha: @Angamanga Thats a good catch. I've just tried it rignt now. Let me see if I can introduce a new property to keep track of the last successfully uploaded key. Or I could simply refine the method instead by removing the if statement

apps/mobile-mzima-client/src/app/profile/information/information.page.ts
```diff
@@ -179,9 +208,197 @@ export class InformationPage {
...
+    this.form.disable();
+    console.log('Starting update process');
+
+    try {
+      let photoUpdated = false;
+      let formUpdated = false;
+
+      if (this.isPhotoChanged) {
+        if (this.selectedFileKey && this.selectedCaption) {
+          await this.uploadPhoto(this.selectedFileKey, this.selectedCaption);
```
Angamanga: I don't think you need the if-statement on row 371 since the filekey always seems to be the same (if I haven't missed anything). For the caption, is that used anywhere? If not, you could just use an empty string or something, if the API requires it.

MMwandigha: @Angamanga Alright Ive removed the unnecessary emission for the caption, the default 'image_upload' caption will now be used instead of emitting the name as the caption

ushahidlee (approved):
Looks good and works well! :+1: :partying_face:

PR #1155, opened 2024-07-09
USH-815 - Inform the user they can't login because the app is offline.

ushahidlee:
**Issue:**

If the user tries to login to the mobile app, but there is no network connectivity, there is nothing to inform the user that this is why they cant login.

**Ticket:**

https://linear.app/ushahidi/issue/USH-815/mobile-offline-lets-tell-the-user-they-cant-login-because-they-are

**Solution:**

Added a toast message that pops up when entering the login screen informing the user of their lack of connectivity. Also rearranged a few networking pieces that were causing silent exceptions.

**Testing:**

1. Open app
2. Select deployment
3. Enter flight mode (or disconnect somehow)
4. Go to login
5. Notice toast message


Angamanga (approved):
🌻 🌻

PR #1159, opened 2024-07-11
Displaying categories when looking at a post in mobile

Angamanga:
Displays categories when looking at a post in mobile.

To test:
- Add a post with a category-field, with categories checked
- Look at the post
- [ ] Categories are visible!!


ushahidlee (approved):
Clean and works well!

PR #1170, opened 2024-07-19
Ush 1417 & 1267 - Questions in the mobile app

ushahidlee:
**Issues:**

This PR solves 2 issues in the mobile app:

- Have survey questions on top of the text field box instead of inside the text field box
- If a survey is created with a long-text field name, the entire text of the field name is not displayed in full on the mobile app but rather truncated with 3 ellipses

**Solution:**

The solution is to move all the field labels to sit above the inputs. There were also a few style changes to match the design ticket, but maybe this needs revisiting as it feels a bit inconsistent.

**Tickets:**

https://linear.app/ushahidi/issue/USH-1417/have-survey-questions-on-top-of-the-text-field-box-instead-of-inside
https://linear.app/ushahidi/issue/USH-1267/long-survey-questions-are-cut-off-in-the-mobile-app

**Testing:**

1. Using the mobile app create a post
2. Notice all the field labels (questions) are above the field inputs.

### DIFF
--- a/apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.html
+++ b/apps/mobile-mzima-client/src/app/post/post-edit/post-edit.page.html
@@ -72,10 +72,10 @@
             *ngIf="field.input === 'text' && (field.type === 'title' || field.type === 'varchar')"
           >
             <div class="form-control--section">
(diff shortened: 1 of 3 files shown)


Angamanga (approved):
Neat 🙌 . Looks good!

Issue #1161, opened 2024-07-15
Recent changes broke posts tests. 


theobadha:
The following tests fail consistently
Automated tests for posts:
- Delete posts
- Verify posts details
fix these tests.

PR #1171, opened 2024-07-22
E2e fix broken posts tests

theobadha:
the broken posts tests proved a bit too flaky after several attempts to fix them. 
I will quiet them down instead, until we can make changes to the platform that will stabilize these tests.


theobadha:
closes #1161

PR #1174, opened 2024-07-23
[USH-1376] Notify user that switching deployments would log them out - FE

MMwandigha:
When a user clicks on a deployment different from the one they're currently logged in to, an alert message displays, warning them that they will be logged out of the current one

This functionality checks if the user is logged in first to show the alert. If they are not logged in, they can easily switch to other deployments without the alert


ushahidlee (approved):
All good :+1:

PR #1175, opened 2024-07-29
[Accessibility] Fixed Keyboard accessibility and operability of collections

### LINKED ISSUE #4984: [Accessibility] Keyboard accessibility and operability of collections
# Summary

## Describe the bug
The existing list of collections is not focusable and operable using the keyboard. When the user presses the Tab key, the focus jumps directly to the Edit and Delete buttons, without the ability to focus on and select individual collections.

## To Reproduce
1. Navigate to the page  that displays the list of collections.
2. Try to navigate through the collections using the Tab key.
3. Observe that the focus jumps directly to the Edit and Delete buttons, without the ability to focus on and select individual collections.

## Expected behavior
When the user presses the Tab key, the focus should first move to the individual collections, allowing the user to select or interact with them. After focusing on a collection, the focus should then move to the Edit and Delete buttons.

## How to solve
1. Ensure that each individual collection in the list is focusable and operable using the keyboard.
2. Implement keyboard event handling to manage the focus flow, ensuring that the focus moves from the collections to the Edit and Delete buttons (JavaScript code would be helpful here, since I have evaluated the code so far and looks like js would be easier to use).

## Screenshots

https://github.com/user-attachments/assets/549ef678-c8a7-4fae-a8df-67c7f02a2c0a

kivuvarosekivuvan:
### Description
- This pull request fixes: https://github.com/ushahidi/platform/issues/4984

### Changes
1. Added `tabindex="0"` to the collection items in the list to make them focusable and operable using the keyboard.
2. Implemented keyboard event handling to manage the focus flow:
   - Added an `(keydown)` event handler to the collection list.
   - In the event handler, checked for the "Enter" key press.
   - If the "Enter" key was pressed, performed the appropriate action (navigate to the collection, edit the collection) based on the target element.

### Screenshots

Before:
The collections were not focusable using the keyboard. Pressing the Tab key would jump directly to the Edit and Delete buttons.

https://github.com/user-attachments/assets/7f9d0e25-fabb-498e-b5a9-e0e3720424cc

After:
The collections are now focusable using the keyboard. Pressing the Tab key moves the focus through the individual collections, and then to the Edit and Delete buttons.

https://github.com/user-attachments/assets/e1ecad8b-9140-496b-a6c7-08dd84c07bbd

### Testing
- Verified the functionality of the Edit and Delete buttons is still working correctly.
- Tested with a screen reader to ensure the keyboard accessibility of the collections.
- Manually inspected the collections to ensure the visual design remains consistent.


apps/web-mzima-client/src/app/shared/components/collections/collections.component.ts
```diff
@@ -82,6 +82,17 @@ export class CollectionsComponent extends BaseComponent implements OnInit {
     this.featuredEnabled = permissions ? permissions.split(',').includes('Manage Posts') : false;
   }
 
+  @HostListener('keydown', ['$event'])
+  onKeydown(event: KeyboardEvent, collection: CollectionResult) {
+    if (event.key === 'Enter' && event.target instanceof HTMLElement) {
```
Ifycode: This is not the review comment for the PR, just me appreciating the "work of art" on the concept and approach generally.
Hmm... This actually looks like some good code you can save somewhere for later @kivuvarosekivuvan
#
> **Extra note:** You may also want to involve the event keycode of the enter key (or which ever key you are doing something like this for).

apps/web-mzima-client/src/app/shared/components/collections/collections.component.html
```diff
@@ -89,6 +89,8 @@
           [checked]="isPostInCollection(item)"
           (checkedChange)="onCheckChange($event, item)"
           [data-qa]="'collection-item'"
+          tabindex="0"
+          (keydown)="onKeydown($event, item)"
```
Ifycode: Now to the review comment for the PR @kivuvarosekivuvan 
Since clicking on the collection cards takes you to some page and closes the modal, logically, it feels like it should be a link. 

Here's my suggestion: Maybe you want to go into/inside the app-collection-item component's html and wrap the code in there with a link instead.

Plus we'll be able to avoid the use of `tabIndex` that way, and also we don't need any or so much javascript for it to cooperate.

And then that it's a link may even also be good for tools like screen readers. since it will announce it that it's a link to some other page/destination.

> It's better to stick with semantic elements

Ifycode (changes requested):

apps/web-mzima-client/src/app/shared/components/collections/collections.component.ts
kivuvarosekivuvan: @Ifycode I understood and read about keycodes(for our case we'll be using code 13). However, I'm wondering since this works at the moment without it, why do you recommend we add it ( more of the impact it'll make to the existing performance )

**Asking to understand what it'll do before implementing . Thanks😊**

.vscode/settings.json
```diff
@@ -0,0 +1 @@
+{}
```
Ifycode: It seems you pushed your VS code settings file and folder along with your change. Please remove it so that it does not show up among files changed @kivuvarosekivuvan 
#
> This is something you can always check for your self in the file changes here: https://github.com/ushahidi/platform-client-mzima/pull/1175/files

Ifycode (changes requested):

apps/web-mzima-client/src/app/shared/components/collections/collections.component.html
```diff
@@ -77,20 +77,24 @@
...
-        </app-collection-item>
+          <app-collection-item
+            [collection]="item"
+            [actions]="item.my_collection || isManageCollections"
+            (edit)="editCollection(item, $event)"
+            (delete)="deleteCollection(item, $event)"
+            [selectable]="post && !!post.id"
+            [checked]="isPostInCollection(item)"
+          >
+          </app-collection-item>
```
Ifycode, line 96: Still checking. Will let you know by around noon when I come back to it.

Ifycode (changes requested):

.vscode/settings.json
kivuvarosekivuvan: @Ifycode Done ✅

apps/web-mzima-client/src/app/shared/components/collections/collections.component.html
kivuvarosekivuvan: Okay @Ifycode

Ifycode: This solution works @kivuvarosekivuvan Nice work
The focus of the top cards are hidden. You can help fix that.

#

#### Screenshot
Screenshot 2024-08-09 at 14 46 59

Ifycode (changes requested):

kivuvarosekivuvan:
@Ifycode The focus visibility for the top cards is fixed, please double check whether is okay on your side. Thanks

Ifycode (approved):
🎉

PR #1177, opened 2024-08-02
Ush 1458 - Handle unknown deployment gracefully

ushahidlee:
**Issue:**

On ushahidi.io, when a user goes to a deployment that doesnt exist or has been deleted, the multi-site handling does not prevent the frontend from continually trying the backend, and the user does not receive any information about what is happening.

**Solution:**

When the frontend receives the "No Deployment Found" response from the backend, it immediately blocks all subsequent backend requests and puts up an alert to the user. The alert allows the user to choose to create a new deployment or exit the site.

**Testing Plan:**

1. Go to a non existent deployment, eg. thisisntadeployment.ushahidi.io
2. Check that all backend requests are halted.
3. Note the alert dialog
4. Click "create new deployment" and be redirected to ushahidi.io/create, or click "exit" and be taken away from the site.

### DIFF
--- a/apps/web-mzima-client/src/app/app.component.ts
+++ b/apps/web-mzima-client/src/app/app.component.ts
@@ -7,6 +7,7 @@ import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { TranslateService } from '@ngx-translate/core';
 import {
   BreakpointService,
+  ConfirmModalService,
   EnvService,
   EventBusService,
   EventType,
@@ -50,10 +51,17 @@ export class AppComponent extends BaseComponent implements OnInit {
     private env: EnvService,
     private gtm: GtmTrackingService,
     private intercom: Intercom,
+    private confirmModalService: ConfirmModalService,
   ) {
     super(sessionService, breakpointService);
     this.checkDesktop();
 
+    if (!this.sessionService.hasSiteConfiguration()) {
+      this.confirmModalService.open({
+        title: 'No Deployment Found',
+        description: 'No Deployment Found',
+      });
+    }
     this.selectedLanguage$ = this.languageService.selectedLanguage$.pipe(untilDestroyed(this));
 
     this.loaderService.isActive$.pipe(untilDestroyed(this)).subscribe({
--- a/apps/web-mzima-client/src/app/app.component.ts
+++ b/apps/web-mzima-client/src/app/app.component.ts
@@ -70,7 +70,7 @@ export class AppComponent extends BaseComponent implements OnInit {
           if (confirm) {
             window.location.href = 'https://ushahidi.io/create';
           } else {
-            window.location.href = 'https://www.ushahidi.com/';
+            window.location.href = 'https://www.youtube.com/watch?v=dQw4w9WgXcQ';
           }
         });
     }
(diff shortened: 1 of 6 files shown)


Ifycode (commented):
Hi @ushahidlee I need clarity on "multi-site handling" which is mentioned in the PR description. I will be in your DM...

Ifycode (changes requested):
I don't know if I should be requesting changes or just commenting on the PR, but I've sent a video to you to explain it better.

Here's the summary: The part where it takes you away from the ushahidi deployment is funny. I'll explain below.

I feel that user knows what they are doing so they deliberately tried to access the Ushahidi deployment. I don't think it's necessary to lead them away from the deployment, since it doesn't look like they wanted to be away from it in the first place.

Where this actually confuses me more, is when I opened up a new tab/window in the browser to test it. And I was led away by the close and exit button to a dancing video on youTube. It felt like there was a security problem with my laptop honestly. More like: "Have I been hacked?" or "is this a scam?" 🤣
Don't mind my laughter but it took me a while to understand what was going on. And I was like: "Oh..." 😲
Even when I am on an already opened browser tab/window it still felt awkward. 

> I realise this may be a product-related concern, but just still dropping it here.

apps/web-mzima-client/src/app/app.component.ts
```diff
@@ -50,10 +51,29 @@ export class AppComponent extends BaseComponent implements OnInit {
...
+          description:
+            'Sorry. That deployment does not exist. Would you like to create a new deployment?',
+          confirmButtonText: 'Create New Deployment',
+          cancelButtonText: 'Exit',
+        })
+        .then((confirm) => {
+          if (confirm) {
+            window.location.href = 'https://ushahidi.io/create';
+          } else {
+            window.location.href = 'https://www.youtube.com/watch?v=dQw4w9WgXcQ';
```
Ifycode, line 73: I see our dancing video here, is this just for testing purpose?

apps/web-mzima-client/src/app/app.component.ts
```diff
@@ -50,10 +51,29 @@ export class AppComponent extends BaseComponent implements OnInit {
...
+    private confirmModalService: ConfirmModalService,
   ) {
     super(sessionService, breakpointService);
     this.checkDesktop();
 
+    if (!this.sessionService.hasSiteConfiguration()) {
+      this.loaderService.hide();
+      this.confirmModalService
+        .open({
+          title: 'Deployment does not exist.',
```
Ifycode, line 63: Nice, but the close Icon on the modal doesn't show up though.

apps/web-mzima-client/src/app/map/map.component.ts
```diff
@@ -77,58 +77,62 @@ export class MapComponent extends MainViewComponent implements OnInit {
...
+      this.reInitParams();
+      this.route.params.subscribe(() => {
+        this.initCollection();
+      });
+      this.initFilterListener();
+      this.mapConfig = this.sessionService.getMapConfigurations();
+
+      const currentLayer =
+        mapHelper.getMapLayers().baselayers[this.mapConfig.default_view!.baselayer];
+
```
Ifycode, line 96: I see some things about map around this area of the code... And how it looks on the UI.

Wouldn't it just be better to approach this like it's done in private deployments?
We usually have no business with the map or any other thing going on, on the UI.
Plus the background behind the modal will be cleaner.

#
#### screenshot with private deployment example
Screenshot 2024-08-06 at 04 47 57

#

> Again, probably a product and/or design related concern...

> But do let me know if I'm over thinking the it, or this is beyond the scope of this PR.

Ifycode (changes requested):

PR #1179, opened 2024-08-07
[USH-1447] In Data View, post details don't change to match selected post

Ifycode:
Add description when done.


Angamanga (approved):
Nice! Works well!

PR #1184, opened 2024-08-13
Ush 1299 - Upgrade to Capacitor 6 (Android 14)

ushahidlee:
**Issue:**

The google play store requires a minimum SDK version of 34 (Android 14) from August 31. 

**Solution:**

To achieve this minimum, we have to upgrade Capacitor to version 6. Also, Android 14 does not require the camera to check for permissions before use (and Capacitor/Camera currently has a bug where this blocks app execution https://github.com/ionic-team/capacitor-plugins/issues/2140)

**Testing:**

As this is a major upgrade of dependencies, both builds of the app will need to be tested (iOS and Android) across all functionality with special focus on native functions:
- Camera use
- Keyboard use
- Sharing


Angamanga:
Looks good. I tested on iOS only and everything seems to be working as it should there. My Android-phone is hiding so could not double-check there right now.

Angamanga (approved):
📱 ☎️  All good!

PR #1186, opened 2024-08-13
Always putting other-option last when creating a survey

Angamanga:
This PR moves the other-option last in the options-array before saving a survey.

To test:
- Create a survey with a radio-button and checkbox-field.
- Add the other option
- Add a new option after adding the other option
- [x] the other option should always be at the bottom of the list of options
- Save the survey
- Add a post to the survey
- [x] the other option should always be at the bottom of the list of options

Note: Surveys with the other-options already configured will not have this change since it is already saved. Test on a new survey or a new field.


ushahidlee (approved):
tidy!

PR #1187, opened 2024-08-14
Activating languages in platform-web

Angamanga:
Activating languages missing in the platform-web client (USH-1009). The languages activated was previously available, and that was causing the bug seen in USH-1455. 
To test:
- In map or data-view, the language dropdown should be visible
- Click on the dropdown and take note of the languages. 
- [x] Following languages are new and should be available: Italian, Albanian, Haitian, Indonesian, Swahili, Vietnamese
- [x] Flags should be available for all languages
- Select one of the new languages
- [x] The platform should be translated to the new language
- [x] Locate at the language-switch dropdown, it should be available in the UI
- Login
- Go to settings->General and click on the language-dropdown
- [x] The new languages should be available there as well
- Select one of the new languages
- [x] The platform should be translated to the new language


ushahidlee (approved):
The dropdown has been conquered!

PR #1188, opened 2024-08-14
[USH-826] Disruptive pop-up notifications(1)

MMwandigha:
The toast message used to appear every time the app was launched.
This solution addresses this issue by only displaying the toast messages when the internet is disconnected and reconnected while the app is still in use.


ushahidlee (approved):
Works nicely

PR #1191, opened 2024-08-14
Ush 1458 - Handle unknown deployment gracefully

ushahidlee:
Issue:

On ushahidi.io, when a user goes to a deployment that doesnt exist or has been deleted, the multi-site handling does not prevent the frontend from continually trying the backend, and the user does not receive any information about what is happening.

Solution:

When the frontend receives the "No Deployment Found" response from the backend, it immediately blocks all subsequent backend requests and puts up an alert to the user. The alert allows the user to choose to create a new deployment or exit the site.

Testing Plan:

Go to a non existent deployment, eg. thisisntadeployment.ushahidi.io
Check that all backend requests are halted.
Note the alert dialog
Click "create new deployment" and be redirected to ushahidi.io/create, or click "exit" and be taken away from the site.

### DIFF
--- a/apps/web-mzima-client/src/app/app-routing.module.ts
+++ b/apps/web-mzima-client/src/app/app-routing.module.ts
@@ -3,6 +3,7 @@ import { RouterModule, Routes, TitleStrategy } from '@angular/router';
 import {
   HostGuard,
   AccessDeniedGuard,
+  DeploymentFoundGuard,
   ResetTokenGuard,
   AccessAllowGuard,
   RedirectGuard,
@@ -13,6 +14,7 @@ import { AccessDeniedComponent } from './shared/components/access-denied/access-
 import { PostNotFoundComponent } from './post/post-not-found/post-not-found.component';
 import { PostNotAllowedComponent } from './post/post-not-allowed/post-not-allowed.component';
 import { RedirectByPostIdGuard } from './core/guards/redirect.post-id.guard';
+import { DeploymentNotFoundComponent } from './shared/components/deployment-not-found/deployment-not-found.component';
 
 const routes: Routes = [
   {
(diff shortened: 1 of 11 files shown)


Ifycode (commented):
I think I will pass the review over to @Angamanga 
I'll just sit this one out, and learn from how the review goes @ushahidlee 😄

#

Referencing previous discussion thread: https://github.com/ushahidi/platform-client-mzima/pull/1177#pullrequestreview-2220198224

#
PS: The close button icon still doesn't show, both on chrome and safari.

apps/web-mzima-client/src/app/app-routing.module.ts
```diff
@@ -95,7 +106,7 @@ const routes: Routes = [
   {
     path: 'post',
     loadChildren: () => import('./post/post.module').then((m) => m.PostModule),
-    canActivate: [AccessDeniedGuard],
+    canActivate: [AccessDeniedGuard, DeploymentFoundGuard],
     data: {
```
Angamanga, line 110: I like this part for some reason, its clear whats going on :D

apps/web-mzima-client/src/app/core/guards/access-allow.guard.ts
```diff
@@ -8,14 +8,13 @@ import { SessionService } from '@services';
...
-  canActivate(): boolean {
+  canActivate() {
     const access: boolean = this.service.accessToSite;
 
-    if (access) {
-      this.router.navigate([`/map`]);
-    } else {
+    if (!access) {
       return true;
     }
```
Angamanga, line 16: I think I get confused with the naming here, why is it returning true when there is !access? Is this correct? It works so its probably correct, just curious :)

Angamanga (approved):
I have some questions in the code but that is out of my own confusion, this works fine and is a neat addition :). Maybe we could brand the popup a little but that is for another time I think...

PR #1193, opened 2024-08-20
Updating order of languages

Angamanga:
Reordering language-file
To test:
- Check the language dropdown
- [ ] The languages are in alphabetic order


ushahidlee (approved):
Looks good but English stays at the top of the list.

PR #1195, opened 2024-08-21
Removing duplicated, unused language-strings

Angamanga:
Removing some strings from en.json, that had duplicates and also was unused in the web-client code.


tuxpiper (approved):
👍

PR #1214, opened 2024-08-21
[Old PR] Children categories not visible in filters from under Parent Category

Ifycode:
**Fixes the following**
- Children category UI and filters fix for mobile client


Angamanga (changes requested):
We discussed this in Slack @Ifycode. While this part works, the "dropdown" function from the designs are missing, add that piece as well and this is good to go 👍

Ifycode:
Sorry for the confusion @Angamanga, I made a mistake with my local branches and pushed the wrong thing to the wrong place. I'll be creating another pull request for this.

PR #1274, opened 2024-08-23
Ush 1458 - Deployment not Found (with extra logic)

ushahidlee:
**Issue:**

The previous fix for the deployment not found issue (https://linear.app/ushahidi/issue/USH-1458/unhandled-deployment-error-client-stuck-in-loading-loop-on-subdomain) sometimes encountered a timing issue that prevented it from correctly assessing whether the deployment was not found, or just had not yet been loaded. This was due to the guard being executed faster than the backend could provide the data.

**Solution:**

I have added an observable to the session service that is updated only after config has been loaded or has failed to do so. The route guard that is checking to see if the deployment is found, will wait until this observable presents success or failure before running its checks.

**Testing:**

1. Open site with proper backend and the site will load.
2. Open site with bad url or no backend and you will see "Deployment not found" page.


Angamanga (approved):
The addition makes sense + it works fine!

PR #1332, opened 2024-09-03
[USH-1473] [Accessibility] Localization Issue in Search Form Component

Ifycode:
**Fix**
- Translate filters related text


Angamanga (approved):
Thanks for this, approved!

PR #1333, opened 2024-09-04
[USH-1507] The language dropdown icon disappears when the language abbreviation is longer than 2 letters.

MMwandigha:
CSS Adjustments to language dropdown


Angamanga (approved):
Very neat! Approved!

PR #1359, opened 2024-09-06
[USH-1487] Data export as a CSV file is not updated immediately in the Export History

MMwandigha:
**Describe the bug**
Data export as a CSV file is not updated immediately in the Export History but rather after refreshing the page.

**Approach**
Calling the `loadExportJobs()` method every time a user exports 'all data' or 'selected fields' so as to refresh the list without requiring a page reload/refresh.


Angamanga (approved):
Well done 💪 ! Approved!

PR #1360, opened 2024-09-09
[USH-1504] Clicking on post preview lags for a short while before the details show up

Ifycode:
**Fix**
- Adds a loader to the post details UI to prevent the said delay on UI, due to wait on the API to fetch a different post onto the screen.


Ifycode:
#### Extra note on the review
This is just for post details. I wanted to know if this is fine before I proceed to do Post Edit (although post edit skeleton loader will look slightly different according to the structure of the elements on the edit route).

> If this is fine, I can just create a separate ticket for post edit and work on that separately.

Angamanga (approved):
Works really well, approved!

PR #1361, opened 2024-09-11
[USH-1200] Web Client: No longer possible to add saved filters

Ifycode:
#### Fix
- Add limit:1000 to saved searches to show all savedsearches for now


Angamanga (approved):
Works fine 🌞

PR #1362, opened 2024-09-15
[USH-1402] Collections created with Only Me visibility cannot have the visibility changed

MMwandigha:
**Issue Description**
Creating a collection with only-me visibility doesn't allow the user to change it to 'everyone' or 'specific_roles' during editing.


Angamanga (approved):
Works fine now! Thank you 🙏

PR #1372, opened 2024-10-04
USH-1267 - Fixed radio option wrapping bug

ushahidlee:
**Issue:**
The radio button options in the mobile app would be truncated with ... added if they exceeded a certain length (device dependent).

**Solution:**
This fix allows the radio button options to wrap within the radio selection div.

**Testing:**
Long radio button options wrap effectively.


MMwandigha (approved):
Works perfectly. Well done!🎉🎉

PR #1373, opened 2024-10-04
[USH-1572] Change UI to help users remember to mark translation task as complete

Ifycode:


Angamanga (approved):
Nice change with the switch to green when the task is completed 🙏 Approved!

PR #1374, opened 2024-10-07
Hiding conversation-window for messages from sislog

Angamanga:
Hiding reply-window for messages from sislog temporarily


ushahidlee (approved):
Looks good :sunglasses:

PR #1376, opened 2024-10-09
Removing comment in html

Angamanga:
Removing a comment in the html that I used the wrong formatting for so its now visible in the platform 🙈


ushahidlee (approved):
haha :+1:

PR #1378, opened 2024-10-15
Using only-parameters for surveys

Angamanga:
This PR changes to use only-parameters for surveys for:
- Add post-modal
- Survey-filters
- Edit/add surveys in settings (The survey-list)
- Data import

To test: Smoke-test on above functionalities, they should work as before 😬

### DIFF
--- a/apps/web-mzima-client/src/app/post/add-post-modal/add-post-modal.component.ts
+++ b/apps/web-mzima-client/src/app/post/add-post-modal/add-post-modal.component.ts
@@ -1,5 +1,5 @@
 import { Component, OnInit } from '@angular/core';
-import { SurveysService, generalHelpers } from '@mzima-client/sdk';
+import { SurveysService, generalHelpers, apiHelpers } from '@mzima-client/sdk';
 import { Observable, of } from 'rxjs';
 @Component({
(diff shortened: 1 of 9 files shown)


apps/web-mzima-client/src/app/post/add-post-modal/add-post-modal.component.ts
```diff
@@ -1,30 +1,37 @@
...
+  constructor(private surveysService: SurveysService) {}
+  ngOnInit() {
+    this.getSurveys();
+  }
 
-  constructor(private surveysService: SurveysService) {
-    this.getPostAllowedTypes();
+  private getSurveys(): void {
+    this.surveysService
+      .getSurveys('', { only: 'name,color,everyone_can_create,can_create' })
```
ushahidlee: maybe the string literal should be in a constant file?

ushahidlee (approved):
Looks good and works nicely. I left a little suggestion but otherwise :+1:

Angamanga:
@ushahidlee And done! Moved the strings to the apiHelper, I think that makes sense since they will be the same in the mobile-app probably...

PR #1379, opened 2024-10-16
Avoid tossing post to the end of the list after edit/add to collections

Angamanga:
This PR:
- updates all posts in feed-view after a change is made to a post
- fixes a small bug connected to add to collection from detail-mode

To test:
1. Try editing, changing status and add to collection to a post in both detailed view and preview-view
- [ ] Feed view should update
- [ ] The same post should be selected
- [ ] The same post should be at the same position as before the update (not at the bottom of the page)
- [ ] The same information should be on the post-preview card and the post-detailed view

Note: I caved in and updated all posts. The benefit is that we now also update all post-locks after a post is saved which is good. The negative thing is that it reloads everything and that takes some time.

Fixes: USH-1330, USH-1344 and USH-1446


apps/web-mzima-client/src/app/feed/feed.component.ts
```diff
@@ -213,7 +213,7 @@ export class FeedComponent extends MainViewComponent implements OnInit, OnDestro
         this.urlAfterInteractionWithFilters = '';
         //-----------------------------------
 
-        this.activeCard.scrollCountHandler({ task: 'increment' });
+        this.activeCard.scrollCountHandler({ task: 'reset' });
 
         this.masonryUpdateOnModeSwitch({ userEvent: this.userEvent });
```
Angamanga: @Ifycode I may have messed up something else, but I could not see what. This increment was causing the post to not scroll into view after update 🤔

Ifycode: Ok, I see the issue. 

The `feedview_postObj`'s post id in the local storage is the same as the post id in the browser URL (i.e. the post id is stored so that we can target the associated postcard for scrolling).

Postcard is only to scroll to view at the point when `scroll_count` is at number `1`.

The `reset` task removes `scroll_count` property from local storage. So applying reset on navigation end would prevent scroll since there's no scroll count to make reference to, though `feedview_postObj` has the stored post id prepared.

> Comment continued on line `888` below.

apps/web-mzima-client/src/app/feed/feed.component.ts
```diff
@@ -876,11 +884,8 @@ export class FeedComponent extends MainViewComponent implements OnInit, OnDestro
     }
   }
 
-  refreshPost({ id }: PostResult) {
-    this.postsService.getById(id).subscribe((p) => {
-      const updatedPost = _.cloneDeep(p);
-      this.posts = this.posts.map((obj) => (obj.id === updatedPost.id ? updatedPost : obj));
-    });
+  refreshPost() {
+    this.getPosts({ params: this.params });
```
Ifycode, line 888: Instead, all you need to do is to use task of `startCount` to control scroll back to number `1` after the refresh (example in line `334` window resize function).

Ifycode (changes requested):

apps/web-mzima-client/src/app/feed/feed.component.ts
MMwandigha: Good catch @Ifycode !

Angamanga:
@Ifycode Updated with your suggestion 🫡

apps/web-mzima-client/src/app/feed/feed.component.ts
Ifycode: What I mean is to call the scroll function/method under this line `this.getPosts({ params: this.params });` in the reset function block, and then using `startCount` task here.
Instead of changing the task for the one in the router navigation end @Angamanga

Angamanga: Ahhh! I see it now 🙏

Angamanga:
@Ifycode Now I think its done :D 🙏 Thank you for your suggestions!

Ifycode (approved):
🎉
#
> PS: hoping we find a fix for the total refresh happening in many places subsequently. Until then this is 👍🏼
> Thank you @Angamanga

PR #1380, opened 2024-10-18
[USH-1443] First page of posts not seen on navigating back to My posts page from viewing a post

MMwandigha:
**Ticket Issue**

When a user selects any posts from ‘my posts’ and then navigates back to the list of posts, not all posts will show up from the top, only posts that belong to the current pagination state onwards.

**Approach**

Introducing two new parameters `reset` and `reloadAll`

- `reset `resets the pagination and clears the post list before fetching new posts from the beginning
- `reload` fetches all posts at once while bypassing pagination. After fetching, the original pagination values are then restored

### DIFF
--- a/apps/mobile-mzima-client/src/app/profile/posts/posts.page.ts
+++ b/apps/mobile-mzima-client/src/app/profile/posts/posts.page.ts
@@ -58,13 +58,13 @@ export class PostsPage {
         if (this.isEditMode) {
           this.editPosts();
         }
-        this.getMyPosts();
+        this.getMyPosts(true);
       },
     });
   }
 
   ionViewWillEnter(): void {
-    this.getMyPosts();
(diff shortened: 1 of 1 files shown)


Angamanga (approved):
I could not reproduce the but on my iPhone and also not in Android studio, but when reading the bug-report and seeing your changes, it does make sense and it works fine everywhere I have tested. So, Approved!!

MMwandigha:
> I could not reproduce the but on my iPhone and also not in Android studio, but when reading the bug-report and seeing your changes, it does make sense and it works fine everywhere I have tested. So, Approved!!

@Angamanga  Great. Thanks for reviewing.
Just to crosscheck (previous build), try to have at least 7 or more posts, then click on the 7th post, and then come back to the list and observe how it behaves. From there, hopefully you will be able to reproduce the bug

PR #1388, opened 2024-10-23
Unhiding language-switcher and adding more languages

Angamanga:
Unhiding the language-switcher in mobile-app + adding more strings and languages

In mobile-app, go to Profile->Languages
- [ ] Czech, Swahili and English should be present
Select a language
- [ ] App should be translated to the selected language.
Note, some strings may still be in English, the translation-threshold is 80% and there are also yet strings that aren't added to Transifex (will create a new ticket for this)


MMwandigha (approved):
Works as expected!🎉

PR #1416, opened 2024-10-31
Ush 1415 - bug fixes

ushahidlee:
**Issues:**

The functionality (described in USH-1415) initially committed had 2 bugs:

- Mime types were not strictly enforced in the frontend.
- An upload that failed due to size, could not be removed.

**Solution:**

This PR contains:
- fixes for both of the above issues.
- much stronger type safety with a class replacing the previous type
- More enums!
- Moves the MediaFile class and its associated Enums into the shared library to be used in mobile.

**Testing:**

USH-1415 functionality but with mime types being enforced and the ability to remove failed uploads.


MMwandigha (approved):
The issues described are now solved. Great job!

tuxpiper (approved):
Looks good to me, I like the re-arrangement 👍

PR #1418, opened 2024-11-04
[USH-1510] Use only parameters for posts in web client

Ifycode:


apps/web-mzima-client/src/app/post/post-preview/post-preview.component.ts
```diff
@@ -44,38 +44,13 @@ export class PostPreviewComponent implements OnInit, OnChanges {
     if (changes['post']) {
       this.allowed_privileges = this.post?.allowed_privileges ?? '';
 
-      // Brute force extraction of the first image found in a post.
-      // Ugly approach, but it stops as soon as it finds an image.
-      // TODO: Optimise
-      if (this.post.post_content) {
```
Angamanga, line 50: Hehe, finally we got rid of "TODO: Optimise" 🤗

Angamanga (changes requested):
There is one silly side-effect on this change, when trying to publish a post from the post-list, the post_content isn't there so the code cannot properly check that all required tasks are completed. It is happening here:https://github.com/ushahidi/platform-client-mzima/blob/development/apps/web-mzima-client/src/app/post/post-head/post-head.component.ts#L86. I think that in the cases when a user is publishing from the feed-view, we would actually need to get the full post. Maybe you can add a check to see if the full post is there, and if not, get it and do the check? It only applies to when the user clicks on "Publish" in the feed-view, not in detailed view.

Ifycode:
I've added the change, do this work? @Angamanga

Ifycode:
This is ready for review too @Angamanga

apps/web-mzima-client/src/app/feed/feed.component.ts
```diff
@@ -725,9 +730,15 @@ export class FeedComponent extends MainViewComponent implements OnInit, OnDestro
...
-        (post: PostResult) => !postHelpers.isAllRequiredCompleted(post),
-      );
+      const uncompletedPosts: PostResult[] = this.selectedPosts.filter((post: PostResult) => {
+        if (post.post_content) {
+          !postHelpers.isAllRequiredCompleted(post);
+        } else {
+          this.postsService.getById(post.id).subscribe((fetchedPost: PostResult) => {
+            !postHelpers.isAllRequiredCompleted(fetchedPost);
+          });
+        }
```
Angamanga, line 740: Just a tiny thing, you need to return the result of the isAllRequiredCompleted to be ably to save it in the uncompletedPosts array 👍 If you do that, it works fine!

Angamanga (changes requested):
Just a tiny change, see my comment in the code 🙌

Ifycode:
Ready @Angamanga

Angamanga (approved):
Approved 🙌

PR #1419, opened 2024-11-04
Ush 1637 - Allow backend_url be specified in mobile app when not in production

ushahidlee:
**Issue:**

The mobile app is designed to have the deployment configured solely through its frontend and doesnt allow the developer to specify a backend in the env.json file.

**Solution:**

This PR allows the developer to specify the backend_url in the env.json which will be used if the app is NOT in production mode. It will then replace all the saved deployments with this one specified deployment and startup with this deployment selected.

**Testing:**

1. Modify the env.json file with production: false and a fully qualified domain in the backend_url, eg. http://localhost:8080/.
2. Start the app.
3. Note the deployment is now the only one specified in the list and is selected.

### DIFF
--- a/apps/mobile-mzima-client/src/app/core/services/env.service.ts
+++ b/apps/mobile-mzima-client/src/app/core/services/env.service.ts
@@ -21,10 +21,15 @@ export class EnvService {
 
   async initEnv(): Promise<EnvConfigInterface> {
     const envy: EnvConfigInterface = await fetch('./env.json').then((res) => res.json());
-    envy['backend_url'] = null;
-    if (this.deploymentUrl) {
-      envy.backend_url = this.deploymentUrl;
+    if (envy['production']) {
+      envy['backend_url'] = null;
+      if (this.deploymentUrl) {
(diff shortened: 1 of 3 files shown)


apps/mobile-mzima-client/src/app/core/services/env.service.ts
```diff
@@ -21,10 +22,28 @@ export class EnvService {
...
-    if (this.deploymentUrl) {
-      envy.backend_url = this.deploymentUrl;
+    if (envy['production']) {
+      envy['backend_url'] = null;
+      if (this.deploymentUrl) {
+        envy.backend_url = this.deploymentUrl;
+      }
+    } else {
+      if (envy.backend_url) {
+        const deployment: Deployment = {
```
Ifycode, line 32: Works, thank you @ushahidlee 

> Dropping the other comments in slack thread here: https://ushahidi.slack.com/archives/C03T5716QE5/p1730949990226169?thread_ts=1730721109.145309&cid=C03T5716QE5

Ifycode (approved):

PR #1422, opened 2024-11-07
Ush 1589 - View Multimedia Content (Mobile)

ushahidlee:
**Issues:**

Users should be able to view posts with new multimedia content: images, audio and documents on mobile.

Update: PR ignores editing new multimedia fields on Mobile.

**Solution:**

This PR adds handling for multimedia fields in the view post templates on mobile.

**Testing:**

1. Create a post with multimedia content (on web).
2. Open the deployment with the mobile app.
3. View the post and note all the multimedia content is displayed correctly.
4. Edit the post and note that the multimedia content does not appear and post can be saved.


apps/mobile-mzima-client/src/env.json
```diff
@@ -1,5 +1,5 @@
 {
-  "production": true,
+  "production": false,
```
Angamanga: Should it still be production: false?

Angamanga (approved):
Just a question about the env.json file, but APPROVED 🌮

PR #1428, opened 2024-11-11
[USH-1605] Add question mark in "Which categories should be available:"

MMwandigha:
Adding question mark in 'which categories should be available'


Angamanga (approved):
🌮 🎏

PR #1429, opened 2024-11-12
[USH-1604] When adding a radio button/checkbox field in a survey, indicate it's mandatory

MMwandigha:
An asterisk (*) has been added to notify the users that the field options are required when adding a radio button field and a checkbox field in a survey


Angamanga (approved):
Approved!

PR #1456, opened 2024-11-13
[USH-1609] Add a notification that unsaved changes to a survey will be lost

MMwandigha:
**Adding confirmation modal for unsaved survey changes**

Below are some of the changes that I have made;

- The confirmModals only have cancel(outline) and delete(red) buttons but no action button (yellow), so I have added that to match the Figma designs.
- Added functionality to track the states of the fields/tasks, meaning that the popup will only appear if the fields/tasks values are edited/different from the initial values. Also, the survey will only be saved if the name field is not empty.
- If a user clicks on 'discard changes', the view will navigate back to the list of surveys. If they click 'save changes', the survey will be saved accordingly.
- Localized the strings

How to test;

1. Open Settings>Surveys>Add Survey and fill in the forms. Try to hit the cancel button and confirm if the popup appears and behaves accordingly.
2. Similarly, select any survey, change any form values/fields, hit the cancel button and confirm if the popup appears and behaves accordingly.


Angamanga (approved):
Works fine! But the discard-button isn't red as we discussed yesterday, wasn't it possible? 
Screenshot 2024-11-14 at 17 43 32

MMwandigha:
> Works fine! But the discard-button isn't red as we discussed yesterday, wasn't it possible? Screenshot 2024-11-14 at 17 43 32

@Angamanga I noticed that the cancel button has the `mat-dialog-close="false"` setting, which means that it doesnt have to be defined explicitly in every modal/notification, therefore this cancel button appears in every modal by default. So I decided to utilize it instead.

I can set a condition to display the 'red' color if the button is destructive as shown in the latest commit, let me hear what you think

image
image

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -124,12 +131,37 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
...
+      this.changesMade = true;
+    });
+  }
+
+  private setInitialState(): void {
+    this.initialFormValue = _.cloneDeep(this.form.value);
+    this.initialTasks = _.cloneDeep(this.form.get('tasks')?.value || []);
+    this.changesMade = false;
+  }
+
```
Angamanga, line 155: Neat 🙌

Angamanga (approved):
Nice change, works fine!

PR #1459, opened 2024-11-14
[USH-1539-ALT] Fix for collection editing rights

MMwandigha:
Previously, the users with the role of 'Manage Collections and Saved Searches' permissions would only edit the collections that they created themselves, which is ideally not the way this works on the web platform

This pull request will now allow the users to edit any collection as long as they have the permission.

Here's the thread that prompted these changes to be made;
https://ushahidi.slack.com/archives/C07PW4NME5S/p1730365088343109?thread_ts=1730118226.988359&cid=C07PW4NME5S


Angamanga (approved):
Nice change @MMwandigha and good catch there @Ifycode. Teamwork 🙌

PR #1460, opened 2024-11-15
[USH-1238] "There are no posts yet" - Separate the text for when there are no posts & when no filters match

Ifycode:
#### Solution implemented explained
- Show "There are no posts yet" text on feed view, when default filters is on (i.e. all 6 filter types are in default state) but there are no posts. There are 2 scenarios for this:
  - There are no posts at all in the deployment: You can test this by logging in as an admin for a deployment that has no posts at all.
  - There are posts in the deployment, but none can be viewed by a user due to the permissions they have.
- Show "No posts match the current filters. Please Change or Clear filters to see matching posts" text on feed view, when not default filters (i.e. at least one of the 6 filter types are in default state) and there's no matching post for user's filter selection/combination
- Search by keyword detection and no posts message:
  - Keyword search only (i.e. searching posts by keyword when Default filters is on): Use "No posts match your keyword search. Please change or clear keyword search to see matching posts".
  - Keyword search + filters applied (i.e. searching posts by keyword when not in default filters): Use "No posts match your keyword search + filters. Please change or clear keyword search and/or filters to see matching posts".

### DIFF
--- a/apps/web-mzima-client/src/app/feed/feed.component.html
+++ b/apps/web-mzima-client/src/app/feed/feed.component.html
@@ -211,16 +211,29 @@
       }"
       [attr.data-qa]="'posts'"
     >
-      <!-- No posts 1: When there are no posts at all in the deployment or there no posts because none can be viewed by a user due to the permissions they have  -->
-      <ng-container *ngIf="!posts.length && isDefaultFilters">
-        <span class="posts-empty">{{ 'post.no_posts_yet' | translate }}</span>
-      </ng-container>
-      <!-- No posts 2: When there's no posts matching user's filter selection/combination -->
-      <ng-container *ngIf="!posts.length && !isDefaultFilters">
-        <span class="posts-empty">{{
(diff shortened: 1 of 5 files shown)


Ifycode:
PR now ready for review @Angamanga

Angamanga (approved):
This works good, just a question, what does this change do?

https://github.com/ushahidi/platform-client-mzima/pull/1460/files#diff-832d69db9caa93b602126446276627d56f1d6713e5c8282cf08c67117fab2439R228

Ifycode:
> Pasting answer here too for anyone interested: 

What the change does - `this.clearPostsResults();` is supposed to restore `this.searchQuery` to an empty string on page reload, so that the `localStorage.setItem('USH_searchPostByKeyword', this.searchQuery);` can be correctly set at that point. Otherwise, localstorage still has a previous value even though the search by keyword input field is cleared/empty on reload. No posts message will therefore not display the correct message we intend.

PR #1461, opened 2024-11-15
Adding extra check for form_id and displaying messages in detailed view

Angamanga:
Adding an extra check for form_id to not having to be dependent on post-fields to display the post.

To test:
Go to feed-view
Click on an unstructured post
- [ ] The content should be visible in the post-detailed view
Click on a structured post
- [ ] The content should be visible in the post-detailed view


ushahidlee (approved):
looks good!

PR #1463, opened 2024-11-17
Removing unneccessary calls when opening/editing a post

Angamanga:
This PR introduce a resolver instead of a guard for post-detail/post-edit view. In this way, we can still check the permissions and also reuse the post-object in the components, and in that way removing an unneccessary call to the API that slowed down the opening of the post.
To test:
1. Go to feed-view
2. Click on a post
- [x] The post should open (but faster!!)
3. Click on the pen on the post
- [x] The edit-view should open with the correct info (but faster!!)
4. Test all kinds of actions, changing status, adding to collections etc etc, they should all work and be reflected in the UI
5. Test on small screens and on the map, the modal should open and all functionality should work as before
6. Test accessing the edit-url as a logged out user
- [x] You should get redirected to the "not found" url
7. Test accessing a post under review as a logged out user
- [x] You should get redirected to the "not found" url

Note: While testing this, I found an existing bug in the mapview:
https://linear.app/ushahidi/issue/USH-1677/changing-status-on-a-post-in-map-view-doesnt-update-the-ui-to-reflect I have not fixed that in this PR.


Ifycode:
@Angamanga I did notice something with the UI and it's dependence on the API call. Link to video here: https://ushahidi.slack.com/archives/C03T5716QE5/p1731926348116819

Ifycode (changes requested):
@Angamanga I did notice something with the UI and it's dependence on the API call before it responds to user action. Affects card preview slide out on click and Mode switch in general on feed view.

Link to video here: https://ushahidi.slack.com/archives/C03T5716QE5/p1731926348116819

Angamanga:
@Ifycode @ushahidlee I cannot believe it but here it is, ready for re-review! 👀 👀

Angamanga:
I will create a new ticket for how we deal with the modal, I patched it together, but would do good with some more work.
New ticket for post-details-view on small screens: https://linear.app/ushahidi/issue/USH-1701/rewrite-how-we-handle-post-details-in-feed-view

ushahidlee (approved):
Code is clean and it works great! :+1: :+1: :+1:

Ifycode:
Mummy wa 🙌🏽
Apart from the post details/edit area first going grey on postById load, everything is good!
Wehdone ma.

Ifycode:
I also noticed that the initial time you get a postById, the skeleton loader takes over the whole page, instead of post cards just being on the left.
Though it's been merged I still wanted to drop this about that aspect of the user experience that is gone @Angamanga

Angamanga:
>I also noticed that the initial time you get a postById, the skeleton loader takes over the whole page, instead of post cards just being on the left.
Though it's been merged I still wanted to drop this about that aspect of the user experience that is gone @Angamanga 

Good catch @Ifycode! I'll create a new ticket for that 👍

PR #1465, opened 2024-11-18
[USH-1610] On edit survey screen, show the user the type of field

Ifycode:


Ifycode:
Ready for review @Angamanga 
> Additional note: Using `mzima-dev` as reference some fields in **_older surveys_** don't have the tool tip popup e.g. "Security issues" survey.
> Newer surveys don't have this problem e.g. "Test survey issues further (do not delete)" survey

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.ts
```diff
@@ -356,4 +365,35 @@ export class SurveyTaskComponent implements OnInit, OnChanges {
...
+    ---------------------------------------------*/
+    const fieldsWithOriginalFieldName = fields.map((field: any) => {
+      if (field.type === 'title' || field.type === 'description') {
+        field.label_original = this.translate.instant(`survey.${field.type}`);
+      } else {
+        const sameField = this.helperSurveyFields.filter(
+          (helperField) => field.input === helperField.input && field.type === helperField.type,
+        )[0];
+        //--- Product decision: Change Label if it's a "Select" field-----------------
+        const originalLabel =
```
Angamanga, line 390: I think this comment belongs in a ticket, not in the codebase :)

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.ts
```diff
@@ -356,4 +365,35 @@ export class SurveyTaskComponent implements OnInit, OnChanges {
...
     this.errorFieldChange.emit(value.trim().length === 0);
   }
+
+  public trackHoverEvent(event: Event, field: any) {
+    /* --------------------------------------------
+      Show an info button on survey fields on hover
+    ---------------------------------------------*/
+    this.fieldHover = event.type === 'mouseenter';
+    this.fieldId = field.id;
+  }
```
Angamanga, line 375: This works, with the mouse. Is there a way to make this available also for keyboard-users? 😬 I think perhaps with focus and blur-event + aria-labels? 🤔

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.html
```diff
@@ -83,16 +83,34 @@ <h2>
...
+              <mzima-client-button
+                *ngIf="fieldId === field.id && fieldHover"
+                fill="clear"
+                color="secondary"
+                [iconOnly]="true"
+                class="tooltip-btn"
+                tabindex="-1"
+                matTooltipPosition="right"
+                matTooltip="{{ field.label_original }}"
+              >
```
Angamanga, line 110: We don't handle this in other places so it might be time for a separate ticket for making the tooltip accessible.

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.html
```diff
@@ -83,16 +83,34 @@ <h2>
...
-            <span class="field-item__title" [data-qa]="'field-name'">{{
-              isDefaultLanguageSelected
-                ? field.label
-                : field.translations[selectLanguageCode]?.label || field.label
-            }}</span>
+            <div
+              class="field-item__main-inner"
+              (mouseenter)="trackHoverEvent($event, field)"
+              (mouseleave)="trackHoverEvent($event, field)"
+            >
```
Angamanga, line 90: I think we need to think about the keyboard-users here as well, not just the hovering?

Angamanga (changes requested):
Just a tiny comment for removing a comment. I think for making the tooltip accessible, we could add a new ticket for that.

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.ts
Ifycode: About the comment, the reason for adding it is based on the discussion I had with Tim. The word "select" is still used in other places e.g. the Add field modal popup when adding fields to the survey. But we are only changing the "select" on the hover text.

At the moment Tim, says he'd have to discuss with other team about if it should also change in other places. So whether it eventually changes or not, the comment is just to help remember or point attention to that slight change - A ticket would usually get closed and forgotten (happened many times) @Angamanga

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.html
Ifycode: There is this part too... 👍🏼

apps/web-mzima-client/src/app/settings/surveys/survey-task/survey-task.component.html
Ifycode: This crossed my mind too. One reason why I added the videos about the 5 different options to the ticket, so that people can check it out, see and make comments @Angamanga 

The most accessible looking option is the one where all the icons show up already without having to hover. Just that there's the problem of having too much or much more elements/information on the UI.

 I've tried one or 2 things accessibility wise on this hover option we are going with, non is successful so far.

I will try again, and let you know how it goes:
> This works, with the mouse. Is there a way to make this available also for keyboard-users? 😬 I think perhaps with focus and blur-event + aria-labels? 🤔

Angamanga (approved):
Thank you Mary! 🙌

PR #1466, opened 2024-11-18
[USH-1546] Loading surveys on mobile still shows "Getting data" instead of "Fetching Surveys"

MMwandigha:
Show 'fetching surveys' instead of 'getting data' for loading surveys

**To test;**

- [x] Open the mobile app 
- [x] Click on the + icon to add a new post 
- [x] Observe the text that appears while fetching the list of surveys


Angamanga (approved):
Thank you for fixing this 🙌

PR #1474, opened 2024-11-20
USH-1683 - Posts not submitting for surveys with the old image field

ushahidlee:
**Issue:**

For surveys with the previous image field, posts fail to submit if an image is not uploaded.

**Solution:**

A tiny little fix that presents an empty array, rather than a null value in the field values.

**Ticket:** 

https://linear.app/ushahidi/issue/USH-1683/posts-not-submitting-for-surveys-that-already-had-image-field#comment-03592f34

**Testing:**

1. Create a survey with an old image field.
2. Create a new post of the above survey, but without uploading an image.
3. Submit the post successfully.


Angamanga (approved):
🦖

PR #1478, opened 2024-11-21
USH-1684 - Edit posts with old image fields

ushahidlee:
**Issue:**

When editing posts that have the old image field, but do not have an image uploaded, the user cannot save the edit due to an error around caption being missing.

**Solution:**

This PR adds better sanity checking in the old code, to prevent these errors from occurring. It also fixes some unreported issues around deleting and replacing images.

**Testing:**

1. Create a post with a survey that has the old image field, but do not upload any images.
2. Edit that post, without modifying the image field (no image or caption).
3. It will save changes successfully.


Angamanga (approved):
Woop!

PR #1479, opened 2024-11-21
USH-1686 - Cutoff posts for empty images

ushahidlee:
**Issue:**

Posts that use the old image field, are cutoff when no image was uploaded.

**Solution:**

This PR adds some minor sanity checks to prevent this from happening.

**Testing:**

1. Create a post in a survey that uses the old image field, but importantly this field cannot be the last one in the post. Make sure there are fields filled in after the image field and that the image field itself is empty (no file uploaded).
2. Open the post in detail view.
3. See that all fields are included in the view and the post is not cutoff.


Angamanga (approved):
How did this even work before :D Approved!

PR #1480, opened 2024-11-21
[USH-1628] Use only parameters for collection in web client

Ifycode:


Angamanga (approved):
Approved, but found another bug connected to collections (not caused by this though): https://linear.app/ushahidi/issue/USH-1689/selecting-a-collection-does-not-show-any-posts

Ifycode:
Thank you.

PR #1481, opened 2024-11-22
Hot fix 2024 45 2

tuxpiper:


ushahidlee (approved):
contains all the bits and tests well

PR #1485, opened 2024-11-26
USH-1694 - Editing a post multiple times, that has a legacy image field causes weird issues

ushahidlee:
**Issue:**

When editing a post that has any legacy image fields, the platform sometimes rejects the edit after trying to edit the same record multiple times.

**Solution:**

This issue is due to the inconsistent use of singular and array values with legacy media items. This PR flattens out any legacy media item array values when deciding how to submit a given post.

**Testing:**

1. Create a post with a legacy image field.
2. Edit the post saving the changes (any changes: add image, delete/replace image, update caption and also other fields)
3. Edit the post again.
4. Succeed with both edits.


Angamanga (approved):
🥇

PR #1487, opened 2024-11-26
[USH-1698] The discard-button when deleting a post has disappeared

MMwandigha:
Removed the unnecessary action button and defined two conditions to the ones that existed before

- By default the 'confirm' button is set to `danger,` but if the button isn't destructive, it will display in `primary`
- By default the 'cancel' button is set to `outline`, but if the button is destructive, it will display in `danger`

Developers will now have to define if the buttons are destructive or not in the code so as to conditionally change their appearance eg:
   
```
isCancelDestructive?: true;
isConfirmNotDestructive?: true;
```
How QA can test this

- [ ] Open any confirmation modal in the system to check this
- [ ] Try creating a new survey > insert text on any field > press cancel instead of save changes ...and observe that the confirmation modal is one of the examples that fulfills what was stated above

### DIFF
--- a/apps/web-mzima-client/src/app/core/services/confirm-modal.service.ts
+++ b/apps/web-mzima-client/src/app/core/services/confirm-modal.service.ts
@@ -8,8 +8,8 @@ interface ConfirmModalProps {
   buttonSuccess?: string;
   confirmButtonText?: string;
   cancelButtonText?: string;
-  actionButtonText?: string;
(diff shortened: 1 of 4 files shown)


Angamanga (approved):
Looking good! Thank you!

MMwandigha:
> Looking good! Thank you!

You're welcome!

PR #1492, opened 2024-11-28
[USH-1632] Use only parameters for category in web client

Ifycode:
Add only parameters for categories:
- Settings:
  - Categories (List, creating or editing)
  - Survey "add field modal" (when adding or editing a new survey)
- Map view's categories endpoint usage


Angamanga (approved):
Thank you @Ifycode 🙌

PR #1494, opened 2024-11-28
[USH-1699] Identify unlocalized strings in the web client and add them to en.json

MMwandigha:


apps/web-mzima-client/src/app/activity/bar-chart/bar-chart.component.ts
```diff
@@ -19,18 +20,24 @@ export class BarChartComponent implements OnInit {
...
   public ngOnInit() {
     this.getPostStatistics(this.selectedFilter);
   }
 
+  private initializeFilters() {
+    this.filters = [
+      { displayNane: this.translate.instant('activity.categories'), value: 'tags' },
+      { displayNane: this.translate.instant('activity.survey'), value: 'form' },
+      { displayNane: this.translate.instant('activity.status'), value: 'status' },
+    ];
```
Angamanga, line 38: Is it actually called displayNane? :D I see it was called that before too, maybe we should change it to displayName, or am I missing something important?

Angamanga (approved):
Well done! Approved!

apps/web-mzima-client/src/app/activity/bar-chart/bar-chart.component.ts
MMwandigha: Ah yes, I should change that. I almost thought it was named like that on purpose but I can still change it :)

Angamanga: Actually, when thinking about it, its for the activity-view that is most likely going to change very soon, so lets leave it as it. Its out of scope for this ticket :)

MMwandigha: Aah ok, actually I have already made the changes, but I think we can leave it at that

PR #1497, opened 2024-12-03
[USH-1629] Use only parameters for collection in Mobile

Ifycode:
Fix:
Only parameters for collections in mobile client


Angamanga (approved):
All good!

PR #1501, opened 2024-12-11
[USH-1677 TEST]Changing status on a post in map-view doesn't update the UI to reflect the change

Ifycode:


Ifycode:
Also ready for review, just that this one is not working.

PR #1504, opened 2024-12-16
Using only-parameters for surveys + other performance issue

Angamanga:
This PR 1. use ONLY-parameters for surveys and 2. only loads the connected survey when editing a post.

To test:
For filters:
- Select filters->Surveys
- [ ] The list should display Surveys
- [ ] The request to get the survey should use ONLY parameters 
- [ ] When selecting a filter, only posts matching the filter should appear
For adding/editing posts:
- [ ] When adding a post, the survey-list should appear as usual, but faster
- [ ] When clicking on a survey, there will be a little delay, because the whole survey is then loaded from the API
- [ ] When editing a post, only the survey that the post is connected to is loaded from the API

This PR fixes USH-1513 and USH-800


ushahidlee (approved):
Nice and clean and works great !

PR #1505, opened 2024-12-16
Translations v1

Angamanga:
This PR adds the translations-feature to posts.
To test:
Add a new translation:
- As a logged in user with edit-rights...
- Go to data-view and click on a post 
- Click on the translate-icon (the icon is currently only available in the detailed-view. We can discuss if it should be in feed-view/map-view as well when planning for V2)
- [x] Translations-modal opens
- [x] There is a list of languages to choose between
_- [ ] The original-language for a post is displayed first in the list and is disabled (Waiting for backend-fix for this so not working yet)_
- Click on a language
- [x] An edit-view opens
- Add translations and save
- [x] The window closes and the post is displayed in the translated language
- Click on Change language and add a new translation from the modal
- [x] The new language is displayed
- Click on "View original"
- [x] The original is displayed
View a translation
- As a logged out user, click on a published post that has translations
- Select a language in the dropdown
- [x] The window closes and the post is displayed in the translated language

Above should also work on small screens
I haven't covered all cases of closing and opening the modal but test around and see how it behaves...


ushahidlee (approved):
Code looks fine and the functionality is :fire:

Angamanga:
Note: Waiting with merging until it is possible to save base_language.

PR #1508, opened 2025-01-07
[USH-1706] Onboarding guide stops progressing after 'filtering' modal

MMwandigha:
**Issue Description:**
The onboarding flow stops progressing after displaying the 'Filtering' guide. Following this, a persistent dark overlay appears, preventing any further interaction with the system. The overlay cannot be dismissed unless the user refreshes the page to regain access.

**Approach;**
I uncommented the code that delays the highlight update using setTimeout to ensure the DOM is fully rendered before proceeding. This snippet initially checks if the step is dynamic eg. 'sorting' requires the feed view to display so as to highlight the button

### DIFF
--- a/apps/web-mzima-client/src/app/shared/components/onboarding/onboarding.component.ts
+++ b/apps/web-mzima-client/src/app/shared/components/onboarding/onboarding.component.ts
@@ -57,12 +57,12 @@ export class OnboardingComponent implements AfterViewInit {
           this.router.navigate(['/settings']);
         }
 
-        // if (this.onboardingSteps[data.order].dynamic) {
-        //   setTimeout(() => {
-        //     this.customTourService.updateHighlightedElements();
-        //     this.isHidden = false;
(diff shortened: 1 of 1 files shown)


Angamanga (changes requested):
This works fine! I did some digging in the git-history to try to understand this change, at the same time as this was removed, the code on the lines 100-107 were added. Things has changed along the way and now they are never called, that is probably why this breaks now. You can remove that part (100-107) and then we are good to go 👍

MMwandigha:
> This works fine! I did some digging in the git-history to try to understand this change, at the same time as this was removed, the code on the lines 100-107 were added. Things has changed along the way and now they are never called, that is probably why this breaks now. You can remove that part (100-107) and then we are good to go 👍

Ah I see. I also have just noticed the functionalities are (if not) the same. Noted, I will remove those lines

Angamanga (approved):
All good now 🙌

PR #1509, opened 2025-01-16
[USH-1726] All post-data is not visible (guest mode only) in the popup on the map even if a post is published.

MMwandigha:
Assigning postId and getSurvey method to posts in map view


Angamanga (approved):
All good 🙌

PR #1512, opened 2025-01-17
[USH-1733] Floating action buttons in the map view are slightly offset in iOS app

MMwandigha:
Implementing the `safe area insets` in the CSS file to allow for proper alignment


Angamanga (approved):
Thank you for fixing this 🙌

PR #1517, opened 2025-01-28
[USH-1744] Fix missing localized strings in en.json on mobile app

MMwandigha:
fix for missing localization keys & inner HTML for no_posts_found


Angamanga (approved):
All good! Thank you for fixing 🙌

PR #1518, opened 2025-01-29
[USH-1748] Create survey: checkboxes "other" field add hint place holder missing

Ifycode:
Fixed add option place holder.


Angamanga (approved):
Works fine 🙌 Approved!

PR #1520, opened 2025-02-05
[USH-1756] Post with required image field is not working as expected.

MMwandigha:
Made the validator more robust to handle various cases;
- If the value is an array, it checks whether the array has at least one element (value.length > 0).
- If the value is an object, it checks for the presence of at least one media type (photo, audio, or doc).

Also renamed PhotoRequired to MediaRequired to account for the multiple media types


Angamanga (approved):
Works fine now 🙌

PR #1521, opened 2025-02-05
[USH-1681] Show confirmation when one navigates from editing a survey or a post

Ifycode:
#### Fixes
- Angular confirm modal trigger on browser back button
- Angular confirm modal trigger on (sidebar and settings) router links click
- Default browser confirm modal

#### Other things to note
- `no changes made` snack bar popup if cancel button is clicked when there's no changes made to the form
- `canDeactivate` guard triggers all confirm modal activities, except for the cancel button
- `canDeactivate` guard also uses save information while processing routing

### DIFF
--- a/apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
+++ b/apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
@@ -6,7 +6,6 @@ import { Observable } from 'rxjs';
 
 export interface IDeactivateGuard {
   changesMade: boolean | null;
-  browserBackButtonClicked: boolean;
 }
 
 @Injectable({ providedIn: 'root' })
@@ -17,7 +16,7 @@ export class DeactivateGuardService implements CanDeactivate<IDeactivateGuard> {
   ) {}
 
   canDeactivate(component: IDeactivateGuard): boolean | Promise<boolean> | Observable<boolean> {
-    if (component.changesMade || (component.changesMade && component.browserBackButtonClicked)) {
+    if (component.changesMade) {
       const confirmModalResult = this.openConfirmModal().then((confirmed: boolean) => confirmed);
       return confirmModalResult;
     } else {
--- a/apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
+++ b/apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
@@ -1,0 +1,39 @@
+import { Injectable } from '@angular/core';
+import { CanDeactivate } from '@angular/router';
+import { TranslateService } from '@ngx-translate/core';
+import { ConfirmModalService } from '@services';
+import { Observable } from 'rxjs';
+
+export interface IDeactivateGuard {
+  changesMade: boolean | null;
+  browserBackButtonClicked: boolean;
+}
+
+@Injectable({ providedIn: 'root' })
+export class DeactivateGuardService implements CanDeactivate<IDeactivateGuard> {
+  constructor(
+    private confirmModalService: ConfirmModalService,
+    private translate: TranslateService,
+  ) {}
+
+  canDeactivate(component: IDeactivateGuard): boolean | Promise<boolean> | Observable<boolean> {
+    if (component.changesMade || (component.changesMade && component.browserBackButtonClicked)) {
+      const confirmModalResult = this.openConfirmModal().then((confirmed: boolean) => confirmed);
+      return confirmModalResult;
+    } else {
+      return true;
+    }
+  }
+
+  async openConfirmModal() {
+    const confirmed = await this.confirmModalService.open({
+      title: this.translate.instant('notify.default.discard_changes'),
+      description: this.translate.instant('notify.default.survey_has_not_been_saved'),
+      confirmButtonText: this.translate.instant('notify.survey.discard_changes'),
+      cancelButtonText: this.translate.instant('notify.survey.back_to_editing'),
+      isConfirmNotDestructive: false,
+      isCancelDestructive: false,
+    });
+    return confirmed;
+  }
+}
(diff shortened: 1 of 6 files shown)


apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -344,27 +348,57 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
...
+    const formHasChanged = this.hasChanges();
+    if (formHasChanged)
+      $event.returnValue = 'Form has changed! - Just to let browser popup show up';
+  }
+
+  @HostListener('window:popstate', ['$event']) onPopState() {
+    this.browserBackButtonClicked = true;
+  }
+
+  public async openConfirmModal({
```
Angamanga: Move this whole functionality to the guard, you don't need it in the survey-item component

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -344,27 +348,57 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
...
 
+  public showSnackbar(message: string) {
+    this.snackBar.open(message, this.translate.instant('notify.snackbar.close'), {
+      duration: 3000,
+      horizontalPosition: 'center',
+      verticalPosition: 'bottom',
+    });
+  }
+
   navigateBack() {
```
Angamanga, line 352: Trigger this when clicking on Cancel instead

apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
```diff
@@ -0,0 +1,34 @@
...
+      cancelButtonOrRouteLinkClick:
+        formHasChangedButModalConfirmButtonNotClicked &&
+        !component.browserBackButtonClicked &&
+        !component.submitted,
+      browserBackButtonClick:
+        formHasChangedButModalConfirmButtonNotClicked && component.browserBackButtonClicked,
+    };
+
+    if (guardDetects.cancelButtonOrRouteLinkClick || guardDetects.browserBackButtonClick) {
+      component.openConfirmModal({ navigateBack: component.modalConfirmButtonClicked });
```
Angamanga: You don't need all the tracking of buttons clicked in the guard, just check for changesMade, if its true, open the modal from here (we need to make the language more general so we can use it in many components, but I think that is fine.) You need to add something like this to the canDeactivateGuard: 
```  if (component?.changesMade) {
      return this.openConfirmModal().then((confirmed: any) => confirmed);
    } else {
      return true;
    }
```

And move the openConfirmModal()-function to the actual guard.

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -344,27 +348,57 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
...
+  @HostListener('window:beforeunload', ['$event']) unloadNotification($event: any): void {
+    const formHasChanged = this.hasChanges();
+    if (formHasChanged)
+      $event.returnValue = 'Form has changed! - Just to let browser popup show up';
+  }
+
+  @HostListener('window:popstate', ['$event']) onPopState() {
+    this.browserBackButtonClicked = true;
+  }
+
```
Angamanga: This works fine! 🙌

Angamanga (changes requested):
Some changes requested, see my comments in the code. To simplify the flow, move the confirm-modal to the guard, it will require less checks, and you already have all the info you need in the guard 🙌

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
Ifycode: Ok.

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
Ifycode: I will message you on slack for this.

apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
Ifycode: It wasn't as straight forward as just checking for `changesMade` only for all the different aspects.
The part about moving the confirm modal to the guard is clear. When I get to that point where it's not straightforward, I will be able to explain better on slack. For now, I will do the best I can.

Ifycode:
Let me know if this works @Angamanga

apps/web-mzima-client/src/app/settings/surveys/survey-item/survey-item.component.ts
```diff
@@ -344,25 +345,13 @@ export class SurveyItemComponent extends BaseComponent implements OnInit {
...
-        this.save();
-      } else {
-        this.navigateBack();
-      }
-    } else {
-      this.navigateBack();
-    }
+  @HostListener('window:popstate', ['$event']) onPopState() {
+    this.browserBackButtonClicked = true;
   }
```
Angamanga: Remove this part

apps/web-mzima-client/src/app/core/guards/deactivate-settings-route.guard.ts
```diff
@@ -0,0 +1,39 @@
...
+@Injectable({ providedIn: 'root' })
+export class DeactivateGuardService implements CanDeactivate<IDeactivateGuard> {
+  constructor(
+    private confirmModalService: ConfirmModalService,
+    private translate: TranslateService,
+  ) {}
+
+  canDeactivate(component: IDeactivateGuard): boolean | Promise<boolean> | Observable<boolean> {
+    if (component.changesMade || (component.changesMade && component.browserBackButtonClicked)) {
+      const confirmModalResult = this.openConfirmModal().then((confirmed: boolean) => confirmed);
```
Angamanga, line 20: Its enough to just check component.changesMade 👍

Angamanga (changes requested):
Two changes, then we are good to go 👍

Ifycode:
Done @Angamanga

Angamanga (approved):
Works fine 🙌

PR #1526, opened 2025-02-14
[USH-1734] Saved searches does not apply in mobile app

Ifycode:
#### Fixes
- Apply button fix for mobile saved filters modal
- Add current view if missing in mobile saved filters
- Minor fix: null value fixes


Angamanga (approved):
Thank you 🙌

PR #1530, opened 2025-02-19
Adjustments to make sure translations with media-fields are saved

Angamanga:
The format of media and old image fields had to be adjusted before the translations could be saved. 

To test:
- Translations should be saved successfully for posts with all kinds of fields, especially media-fields


ushahidlee (approved):
Them nasty arrays. Looks good :+1:

PR #1534, opened 2025-02-28
First part of map-library migration

Angamanga:
The first part of moving over to OpenLayers, the map is not complete but...

- The map is displaying
- When clicking on a pin, the popup shows and displays information + all functionality such as edit, change status etc works as before

Since filters and loading-posts are covered by another ticket, you might need to reload or go to data-view and back to map-view to get the pins to load.

Ticket: USH-1763


ushahidlee (approved):
Code looks neat and tidy. :+1:

PR #1535, opened 2025-02-28
[USH-1798] Create and edit surveys: Discard modal shows up on save though surveys are saved

Ifycode:
#### Fixes
- Prevent discard changes modal from showing up on save


Angamanga (approved):
Works now! Thank you 🙏

PR #1540, opened 2025-03-07
Displaying map when editing/viewing posts

Angamanga:
This PR is exchanging Leaflet with OpenLayers when creating, editing and viewing a post:
To test:
1. Add a post to a survey with location:
- [x] Make sure the map is displayed with the selected basemap
- [x] Search for a location, the selected location displays with a pin on the map
- [x] Click on the map, make sure the pin is put in the clicked position
- [x] lat/long fields are updated properly
2. Save the post and open the post from data-view
- [x] The post is shown with the map
- [x] The pin is located on the right place
3. Edit the post
- [ ] Make sure the map is displayed with the selected basemap
- [x] Search for a location, the selected location displays with a pin on the map
- [x] Click on the map, make sure the pin is put in the clicked position
- [x] lat/long fields are updated properly

Note: There is a mismatch when looking at posts on the post-details modal. I have created a new ticket for that because it is likely not connected to openLayers but to something with how the post-modal is handled (needs to be worked on before map is released though!). New ticket: https://linear.app/ushahidi/issue/USH-1808/show-map-when-viewing-post-on-smaller-screens-and-on-map-modal

### DIFF
--- a/apps/web-mzima-client/src/app/map/map.component.ts
+++ b/apps/web-mzima-client/src/app/map/map.component.ts
@@ -78,25 +78,26 @@ export class MapComponent extends MainViewComponent implements OnInit {
   }
 
   initBaseLayers() {
-    const baseLayers = mapHelper
-      .getOpenLayersMapConfig()
-      .filter((layer) => layer.visible)
-      .map(
-        (layer) =>
-          new TileLayer({
-            title: layer.name,
-            type: 'base',
-            visible: this.mapConfig.default_view?.baselayer === layer.code,
-            source: new XYZ({
-              url: layer.url,
-              maxZoom: 'maxZoom' in layer.layerOptions ? layer.layerOptions.maxZoom : undefined,
-            }),
-          } as BaseLayerOptions),
-      );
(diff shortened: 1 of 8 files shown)


ushahidlee (changes requested):
A few small issues:

1. The basemap is not selected by default, meaning the deployment opens with an empty map until you select a basemap using the new selector (!).
2. The map that appears in the create/edit modals does not match the base map of the deployment.
3. When creating or editing a post, the post is always saved with the latlon of Nairobi.

Angamanga:
@ushahidlee Ready for re-review 🎉

Angamanga:
@ushahidlee For this one:
> The map that appears in the create/edit modals does not match the base map of the deployment.

The create/edit map will be the default map set in general settings, not the one the user potentially change to in the map-view.

ushahidlee:
The original bugs are now fixed but I have a couple of other issues:

- If I open a post in data view, and then click on map view there is no map.
- If I edit a post from map view, there is no map in the modal

PR #1541, opened 2025-03-10
[USH-1804] Add Kurdish to language-switcher in web-client

MMwandigha:
- Added Kurdish language to the list of languages (both ku and ku_IQ)
- Added their corresponding flags


Angamanga (approved):
Approved!

PR #1542, opened 2025-03-10
[USH-1799] Web client's date filter is not applied - saved filters

Ifycode:
**Fixes**
- Date filters issue for saved filters where date is applied


Angamanga (approved):
Approved!! 🎉

PR #1543, opened 2025-03-10
[USH-1797] [Multimedia] Rename the default descriptions of the multimedia fields for Web + Mobile

MMwandigha:
Changed the description of the fields from `Allows image(s) to be uploaded to the post` to `Upload one or more image file(s)` . Same for audio and docs.

NB: This default description will only appear for newly created surveys or newly added fields


Angamanga (approved):
All good 🙌

